Issue #10, opened 2020-04-05
Fix trait/perk: Mounted Archer


johnshepardn7117:
Same issue as with disciplinarian, So I will list them off here, unless you want me to do a separate link to each?

 If these don't then I assume, many others don't either.

Trait confirmed not to work.

Mounted Archery trait: allowing any bow to be used on horseman.
Mounted Crossbows: allowing reloading on horseback

johnshepardn7117:
Mod claiming to have fixed the issue: (permission needed)

https://www.nexusmods.com/mountandblade2bannerlord/mods/215?tab=files&file_id=842&nmm=1

ajseavey:
I'm the author of that fix and will be looking into incorporating it into this mod tonight.

Jayshua83:
Is this still happening? Would be really nice to not have so many mods! :)

ajseavey:
It is. Working on getting a proper implementation work. Have one last thing to iron out, I think.

Velken:
Cant use Noble Long Bow on horse with Bow Expert nor with Mounted Archer. (e1.1.0-Beta)

Jayshua83:
> Cant use Noble Long Bow on horse with Bow Expert nor with Mounted Archer. (e1.1.0-Beta)

I can with the mod that fixes those perks!

PR #51, opened 2020-04-07
Fix Mounted Archer, Bow Expert, Crossbow Cavalry, and Crossbow Expert perks.

ajseavey:
Addresses #10 by overriding flags and usability on the hero adding said perks or on loading while the hero has said perks. Based on my solution from https://www.nexusmods.com/mountandblade2bannerlord/mods/215.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ Current Fixes:
 * The Peak Form perk.
 * The Ruler perk.
 * The Reeve perk.
+* The Mounted Archer perk.
+* The Bow Expert perk.
+* The Crossbow Cavalry perk
+* The Crossbow Expert perk
 * Fixes Item Comparison perk-based coloring.
 
 Current Features:
--- a/src/CommunityPatch/Patches/MountedBowPerksPatch.cs
+++ b/src/CommunityPatch/Patches/MountedBowPerksPatch.cs
@@ -1,0 +1,54 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+
+namespace CommunityPatch.Patches {
+
+  internal class MountedBowPerksPatch : IPatch {
+
+    public bool Applied { get; private set; }
+
+    public static readonly List<PerkObject> _perks
+      = new List<PerkObject> { DefaultPerks.Bow.MountedArcher, DefaultPerks.Riding.BowExpert };
+
+    public bool IsApplicable(Game game)
+      // ReSharper disable once CompareOfFloatsByEqualityOperator
+      => game.GameType as Campaign != null
+          && (game.PlayerTroop as CharacterObject)?.HeroObject != null;
+
+    public void Apply(Game game) {
+
+      Hero hero = (game.PlayerTroop as CharacterObject)?.HeroObject;
+
+      if (_perks.Any(hero.GetPerkValue)) PatchWeapons();
+
+      (game.GameType as Campaign).CampaignBehaviorManager?.AddBehavior(new OnMountedBowPerkAddedBehaviour());
+      
+      Applied = true;
+    }
+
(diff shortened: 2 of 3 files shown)


Skau:
Nice! You should update the readme too with the fixes.

ajseavey:
Done and done

Tyler-IN:
The patches look ok at first glance...
However, the IsApplicable bits just check to see if it's a Campaign and there's a Player Troop with a Hero? The implementation should check to see if the patch is "applicable" in the sense that it isn't already fixed or patched by another mod.

Looking at the patches a bit more, what's the deal with the serialization stuff and checking if it's the player hero?

ajseavey:
The `AddNonSerializedListener` is just adding an event listener on the `PerkOpened` (Perk Added) event. Also I'm not too savvy on when that triggers, but it may trigger when you do it for a companion, and since as far as I can tell the changes I'm making are global for the ItemObject I'm changing, at least for now, it seems wiser to restrict it to when the player character gets the perk.

Also, sorry about the `IsApplicable` stuff. I wasn't entirely sure of the intended purpose of that was. I'll fix those conditions soon. I may also look into method patching to maybe change it for only characters with the perk.

Tyler-IN:
So would this affect other game characters?

E.g. your hero can equip bows on horseback, but then suddenly so can the opposing heroes?

ajseavey:
I don't know if the opposing heroes would do so, but as far as I can tell this would make it possible for you to equip you companions with a long bow and a horse and they should be able to use it if you have the perk, as far as I can tell. I'm looking for ways to restrict that based on the character but that also is making me wonder if any regular cavalry units use long bow or crossbow.

ajseavey:
So looking at `Agent.TryToWieldWeaponInSlot` it makes a call to an interface full of `EngineMethod` annotated methods which I'm not sure if those are super internal or what. It's a lead to look into but I'm not sure how big of a rabbithole it'll be.

Skau:
There's also `Agent.EquipWeaponWithNewEntity` where you get the weapon in as an parameter directly. Instead of globally setting the values on all weapons you could just do it on the currently equipped weapon, and "turn it off" again when it's dropped or removed. This would make it so only the hero with the perk will have the effect. By using `Agent` you would have to get the corresponding `Hero` to see the perks. Just my two cents, might be better ways.

ajseavey:
Ideally I could find the logic that uses the flags I'm changing and overwrite that logic but it seems like that might be buried pretty deep.

ajseavey:
It does absolutely seem like `Agent.TryToWieldWeaponInSlot` is what I want, but since it makes a call off to that interface I have yet to figure out what I need to modify to make it do what I want.

Tyler-IN:
I don't see any more changes here yet.

Can you implement IsApplicable like I described above?

I'm going to investigate TryToWieldWeaponInSlot and give some feedback here soon.

ajseavey:
Sorry have been at work.  Will do so when I'm done for the day.

Tyler-IN:
See if you can join us on Discord and we can add you as a contributor to the project.

ajseavey:
I'm on the Mount and Blade discord. Svered#1526

Tyler-IN:
Update on this;

We discussed the craziness of how switching weapons and reloading are done.

Native code and callbacks come into play here.
I'd like to try to have this work for all heroes with the right perks, and none of the other parties.
If the AI doesn't ever use crossbows and longbows on horseback, then we can can go with this, but it might screw with other mods or have some strange edge cases.

ajseavey:
I am working on rewriting it to be more generally correct and hopefully that works. If not this can be a fallback implementation.

Tyler-IN:
Just letting the community know, @ajseavey is still working on this issue and has made good progress identifying a way to fix this directly and splitting up the issues into separate patches.

The major hurdle faced with these is that they call into native code in Game.dll.

ajseavey:
Closing since I've since redesigned how I'm doing this and have seperate PRs for those patches.

Issue #33, opened 2020-04-06
Bannerlord Perk doesn't work


tynakuh:
Should give +2 to party size for each fief owned. (It's personal but should probably be clan leader, as they are the only ones who can own fiefs?)

Tyler-IN:
hey @Skau I tagged you on this, you can remove yourself if you have another focus

Skau:
Np, I'll give it a try

Issue #39, opened 2020-04-06
Man-At-Arms doesn't work 


tynakuh:
Should gain 5 party size with every fief owned.

PR #56, opened 2020-04-07
Bannerlord perk

Skau:
This is a fix for #33 

I split it into two patches - one for the actual party size limit logic and one for the tooltip when you hover over the party size as shown here:

TaleWorlds MountAndBlade Launcher_aLZa4o97mB

I postfix PartyBase.get_PartySizeLimit and PartyBase.get_PartySizeLimitExplainer respectively.
In BannerlordPerkExplainerPatch I reuse the method from BannerlordPerkPatch to make it consisent.

Once again we're in the situation on how we want to interpret TaleWorlds wording: 
*+2 party size for each fief that you own*

I opted to go for the approach where the party leader is chosen as the check if the perk is available. Then it uses the party leader current clan and gets all its settlements. This means that any companion or NPC will get benefits of Bannerlord (if they can get the perk at all?). 
All settlements means that the castle/city will give +2 and the attached villages will provide another +2 each. 
Example:
TaleWorlds MountAndBlade Launcher_9HMZPALbfW

There are 6 settlements owned by the clan, so the party leader with the Bannerlord perk will have an extra +12 to its party size limit.

Does that make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Current Fixes:
 * The Peak Form perk.
 * The Ruler perk.
 * The Reeve perk.
+* The Bannerlord perk.
 * Fixes Item Comparison perk-based coloring.
 
 Current Features:
--- a/src/CommunityPatch/Patches/BannerlordPerkExplainerPatch.cs
+++ b/src/CommunityPatch/Patches/BannerlordPerkExplainerPatch.cs
@@ -1,0 +1,52 @@
+using System.Linq;
+using System.Reflection;
+using System.Security.Cryptography;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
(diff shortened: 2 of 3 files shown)


Skau:
This + the steward ruler/reeve patches can probably be refactored out a bit. They patch very general functions (companion limit and party limit) and they could be reused for in other fixes. For example #39  could be implemented here as well, as it's basically the same.

Tyler-IN:
Since you're using AlreadyPatchedByOthers, it should be fine for other patches to apply to those methods and stack. Since you're using postfix patches that only modify the result, I think this will be fine.

Tyler-IN:
LGTM

Skau:
Yeah. It might best to have each patch separated anyways so they can be disabled individually for example if that's ever needed.

PR #57, opened 2020-04-07
Fixed Man-At-Arms

Skau:
Fix for #39.
Interpreted the same as #56 because the wording is exactly the same.

They stack appropriately.
TaleWorlds MountAndBlade Launcher_p4BcHJ2v3J


Tyler-IN:
LGTM - you are knocking these out pretty quick

Skau:
Sweet!

Issue #38, opened 2020-04-06
Swords as Tribute doesn't work


tynakuh:
Don't gain 10 party size for every Vassal you have.

PR #58, opened 2020-04-07
Steward Swords As Tribute fix

Skau:
Fix for #38 
*+10 party size for every vassal*

It properly stacks with the other party size patches.

It checks if the party leader's clan is the ruling clan of the kingdom and gets the count of all clans, excluding itself. Does that make sense? I think that is the vassals. Aka other clans in a kingdom.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Current Fixes:
 * The Ruler perk.
 * The Reeve perk.
 * The Bannerlord perk.
+* The Swords As Tribute (Steward) perk.
 * Fixes Item Comparison perk-based coloring.
 
 Current Features:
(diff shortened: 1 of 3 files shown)


tynakuh:
LGTM, one thing that isn't taken into account is usage in:
`TaleWorlds.CampaignSystem.SandBox.GameComponents.Party.DefaultPartySizeLimitModel.ShowPartySizeDetail`

But that looks like its pure dev/cheat console related and I'm not sure we should worry about stuff like that for the community patch.

Skau:
>TaleWorlds.CampaignSystem.SandBox.GameComponents.Party.DefaultPartySizeLimitModel.ShowPartySizeDetail

I wasn't even aware of this method. There's definitely a decision that has to be made regarding patching with haromy against modifying the default model.

Issue #30, opened 2020-04-06
Peak Form and Healthy Scout Don't Add Health


Talnar:
Using your latest patch on the latest steam game version, selecting these perks still doesn't add max health to the character.

Tyler-IN:
Yep, that's missing - it needs to be implemented. The patches only fix the declarations, not the addition of health.

PR #59, opened 2020-04-07
Fix Healthy Scout & Peak Form perks, implementation via model wrapping

Tyler-IN:
Fixes https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/issues/30

### DIFF
--- a/src/CommunityPatch/Patches/HealthyScoutPatch.cs
+++ b/src/CommunityPatch/Patches/HealthyScoutPatch.cs
@@ -1,7 +1,15 @@
+using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Helpers;
 using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.SandBox.GameComponents;
 using TaleWorlds.Core;
+using TaleWorlds.Library;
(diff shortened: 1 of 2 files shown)


Tyler-IN:
Yeah if another mod changes out the model it doesn't apply.

Might need to add a fallback patch to handle that scenario, but if a mod is replacing the model, it might not even want to take the default behavior into account...

I also don't know if this even works yet, haven't tested in game.

Skau:
Well, I belive they definitely still want the perk fixes even with custom models. That's what I'm getting at. Swapping out a model breaking a perk fix doesn't feel right.

tynakuh:
I like the approach, but agreed, I'd imagine lots of mods will be going down the custom models approach, which means they're losing out on the perk fixes themselves if they haven't implemented them. I think i'd prefer not checking if the models a custom one myself but I can see arguments for and against that.

Is it possible to add a check for if TaleWorlds fixes this, that way when they put a fix out (assuming they fix it in the default model, seems sensible to me though) we aren't doubling perk effects.

Tyler-IN:
This needs an IL hash check

Tyler-IN:
Prefer #60

PR #60, opened 2020-04-07
Fix Healthy Scout & Peak Form perks, implementation via Harmony patching

Tyler-IN:
Fixes https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/issues/30


Tyler-IN:
tested, fixed

Issue #53, opened 2020-04-07
About the DependedModules: Stop doing this.


bighungryjames:
I recognized that you show contempt on the people who are adding \<DependedModules\> to your mod's Submodules.xml file. I see that you think emptying that will guarantee your mod to stay on the top on the loading list, right after the native modules.
However, as I observed the default launcher's behavior, I saw that leaving the DependedModules entry empty does NOT guarantee that your mod is loaded before any other mods. The launcher only checks if all the modules have their depended modules loaded before them, nothing more. Having no depended modules has nothing to do with the load order between the mods. In order to have your desired behavior correctly, all of the mods on the Bannerlord Nexus should contain your mod in their depended modules list, not the other way around.
You see, in a multi-participant project, it is really critical to use functions as intuitive as possible. Your approach is, with all due respect, downright confusing. Sure, the launcher assures any non-native modules to load after native modules, but that is a safeguard mechanism. Since your mod actually depends on the native modules, it is much more desirable for your mod's DependedModules list to contain the native modules. That's intuitive. That's how it should be working and how everybody else is expecting it to be working.
I do understand that there is no way to persuade all of the thousands of the modders to add your mod on their DependedModules list.  I do see why you tried that approach. However, since it is not working as intended, it has no gains but it loses the intuitivity. Please stop that. Instead, direct the people to use the FixedLauncher Mod which allows the people to change load order manually.
Perhaps the best solution to this is Taleworlds adding a tag in the Submodules reader function such as {{{HighPriority = "yes"}}} or so. However, until that happens, it is best to follow the way things are intended to be in.

Tyler-IN:
```cs
		// Token: 0x0600000E RID: 14 RVA: 0x00002298 File Offset: 0x00000498
		private void LoadSubModules(bool isMultiplayer)
		{
			List<ModuleInfo> list = ModuleInfo.GetModules().Where((ModuleInfo x) => this.IsVisible(isMultiplayer, x)).ToList<ModuleInfo>();
			list.Sort(new LauncherModuleInitialComparer(list.AsReadOnly()));
			for (int i = 0; i < list.Count; i++)
			{
				ModuleInfo moduleInfo = list[i];
				LauncherModuleVM launcherModuleVM = new LauncherModuleVM(moduleInfo, new Action<LauncherModuleVM, int>(this.ChangeLoadingOrderOf), new Action<LauncherModuleVM>(this.ChangeIsSelectedOf));
[...]

PR #19, opened 2020-04-06
Update SubModule.xml

Grandnovus:
Updated SubModule.xml to correct the mod order list on the launcher.

Haven't gone through the rest of the code yet, so not entirely sure if this was done on purpose (regarding the ability to delay mods loading).

PR #62, opened 2020-04-08
Update SubModule.xml

data54:
Added DependedModule for the main game files so it doesn't load before the main game files or randomize load order of the other mods

### DIFF
--- a/SubModule.xml
+++ b/SubModule.xml
@@ -6,7 +6,13 @@
     <SingleplayerModule value="true"/>
     <MultiplayerModule value="true"/>
     <Official value="false"/>
-    <DependedModules/>
+    <DependedModules>
+		<DependedModule Id="Native"/>
+		<DependedModule Id="SandBoxCore"/>
+		<DependedModule Id="Sandbox"/>
+		<DependedModule Id="CustomBattle"/>
+		<DependedModule Id="StoryMode"/>
+	</DependedModules>
     <SubModules>
         <SubModule>
             <Name value="CommunityPatch"/>


Tyler-IN:
Is this the same as https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/pull/19 ?

data54:
Looks like it

data54:
Might still be worth doing even if your mod doesn't require it, looks like not having it can potentially break existing mods that do.
With it:
image
Without:
image
Character Trainer DependedModules node:
image

Tyler-IN:
> it can potentially break existing mods that do.

Can you provide an example of it breaking an existing mod?

Grandnovus:
You saw the same thing I did regarding the mod load list (I was the one that opened #19 ).

However using the fixed launcher mod (https://www.nexusmods.com/mountandblade2bannerlord/mods/205) sorted out that issue.

image

I'm not 100% sure if I launched the game using the set-up above but I don't believe it actually had any negative effects.

Tyler walked me through how the launcher loads mods and in what order after I posted #19 and it shouldn't make a difference 

> [1:39 PM] ̑🅠: so ... it basically just goes by depth then file name
> [1:40 PM] ̑🅠: once the submodules are loaded, the dependencies matter
> [1:40 PM] ̑🅠: native comes first, then anything with no dependencies, then anything with dependencies (incl. official modules)

Tyler-IN:
Linking to #53

data54:
> > it can potentially break existing mods that do.
> 
> Can you provide an example of it breaking an existing mod?
> 
> e.g. Does CharacterTrainer actually break, or does the sort order just look wrong?

It doesn't break character trainer but it seems a bit random on what mod ends up ahead of the main game files, guessing there's a null ref or something in the launcher itself that causes it to break if it comes across a mod that doesn't have it in the xml.  When I originally started adding mods to the game the bannerlord tweaks mod was loading before its dependencies which did break the game, only thing that fixed it was making sure every mod I had contained the DependedModules.

PR #63, opened 2020-04-08
HugsLib like options API

### LINKED ISSUE #61: Suggestion: Mod Settings Framework
Hi, I would like to suggest a mod settings framework that other mods can use.

As an example I would like to recommend Hugslib from the Rimworld Community, as I think it's quite simple and effective to use:

usage example:
https://github.com/Fumblesneeze/RimBetterSpeed/blob/master/Source/BetterSpeed/Mod.cs

source is here:
https://github.com/UnlimitedHugs/RimworldHugsLib/tree/master/Source/Settings

the idea is that you as a user implement a base class where you give it a displayname and then you can add SettingHandle<T> fields to that class and initialize them with whatever selection of controls seems good, e.g. slider, textfield, number, and the framework collects all of these on one mod settings page, generates the UI and calls the SettingsChanged() callbacks when settings have been loaded or changed by the user. Reading the settings is easy since the SettingsHandle<T> will implicitly convert to to T.

Tyler-IN:
Begins effort for https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/issues/61


src/CommunityPatch/CommunityPatchSubModule.cs
```diff
@@ -26,7 +27,7 @@ public partial class CommunityPatchSubModule : MBSubModuleBase {
     internal static readonly LinkedList<Exception> RecordedUnhandledExceptions
       = new LinkedList<Exception>();
 
-    internal static readonly OptionsFile Options = new OptionsFile(nameof(CommunityPatch)+".txt");
+    internal static readonly OptionsStore Options = new OptionsFile(nameof(CommunityPatch) + ".txt");
```
tynakuh, line 30: I'd prefer explicitly enforcing a file type/extension here.   Given they're all toml files as well.

src/CommunityPatch/Options/OptionsStore.cs
```diff
@@ -0,0 +1,17 @@
+namespace CommunityPatch {
+
+  public abstract class OptionsStore {
+
+    public abstract void Save();
+
+    public abstract void Set<T>(string key, T value);
+
+    public abstract void Set<T>(string ns, string key, T value);
```
tynakuh, line 9: Can we name these variables to be clearer for API purposes?

tynakuh (commented):
LGTM to me otherwise once we have a second pair of eyes.

src/CommunityPatch/Options/OptionsStore.cs
Tyler-IN: `ns` to `namespace`?

src/CommunityPatch/CommunityPatchSubModule.cs
Tyler-IN: The other simple key-value-pair bannerlord config files are all .txt files, apart from that there's xml files.

src/CommunityPatch/Options/OptionsStore.cs
tynakuh: Yep

src/CommunityPatch/CommunityPatchSubModule.cs
tynakuh: Ah I see the intention now, resolving.

src/CommunityPatch/Options/OptionsStore.cs
Tyler-IN: `namespace` is a language keyword.

Tyler-IN: Discussed in Discord, will add xml docs to flesh out description of parameters.

Tyler-IN:
> Un-needed. Actually exists in Apply already.

I think at some point we'll be displaying some GUI for patches, applicable patches will need to not show as skipped when applicable unless something went wrong when applying.
In this case, after being applied at campaign start up, it'd show as not applicable and applied, but skipped.

Tyler-IN:
This has been forked off to be made into a separate module.

https://github.com/Tyler-IN/MnB2-Bannerlord-ModOptions

CommunityPatch will add it as a dependecy and use it.

Tyler-IN:
This has references to options that need to be refactored and merged once ModOptions is up and running.

src/CommunityPatch/ContractAnnotationAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+    public ContractAnnotationAttribute([NotNull] string contract, bool forceFullStates) {
+      Contract = contract;
+      ForceFullStates = forceFullStates;
+    }
+
+    [NotNull]
+    public string Contract { get; }
+
+    public bool ForceFullStates { get; }
+
```
tryadelion, line 22: remove uneccessary spaces

src/CommunityPatch/Options/CustomOptionCategoryViewModel.cs
```diff
@@ -0,0 +1,29 @@
...
+      // ReSharper disable once PossibleNullReferenceException
+      var optionsBindingList = (MBBindingList<GenericOptionDataVM>) optionsField.GetValue(this);
+
+      foreach (var option in options)
+        optionsBindingList.Add(option);
+
+      // ReSharper disable once VirtualMemberCallInConstructor // because of course we do...
+      RefreshValues();
+    }
+
```
tryadelion, line 26: remove unnecessary spaces

tryadelion (approved):
Just some unnecessary trailing whitespaces / empty ines.

PR #64, opened 2020-04-08
War Rations fix

### LINKED ISSUE #36: War Rations doesn't work
Doesn't reduce amount of food consumed by 20%

Tyler-IN:
Carrying on where @Skau left off tonight;

before

after

Fixes #36.

### DIFF
--- a/src/CommunityPatch/Patches/WarRationsPerkPatch.cs
+++ b/src/CommunityPatch/Patches/WarRationsPerkPatch.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using HarmonyLib;
 using TaleWorlds.CampaignSystem;
 using TaleWorlds.CampaignSystem.SandBox.GameComponents.Map;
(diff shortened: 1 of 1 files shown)


Tyler-IN:
In case someone (probably TaleWorlds) changes the perk, it'll still take affect according to the expected behavior.

Tyler-IN:
There's a separate branch for options that does handles like HugsLib, e.g. Option<T>

src/CommunityPatch/Patches/WarRationsPerkPatch.cs
```diff
@@ -0,0 +1,95 @@
...
+        // qm is not in party
+        && party.MemberRoster.All(element => element.Character?.HeroObject != qm)
+      )
+        return;
+
+      var perk = ActivePatch._perk;
+      if (!qm.GetPerkValue(perk))
+        return;
+
+      var explainedNumber = new ExplainedNumber(__result, explainer);
```
tynakuh, line 78: I have a feeling this is going to be a common thing we'll have to do, wonder if we should build a helper function

Tyler-IN: Possibly. Copy pasta for now, refactor to helper later.

PR #69, opened 2020-04-08
Perk rewrites

Skau:
Rewrote Swords And Tribute, Bannerlord and Man-At-Arms patches to use the model approach seen in other patches. Reason is 1. to stay consistent and 2. that's most likely where TW would patch them.
I did not rewrite Ruler and Reeve because there are not an appropriate place to do so, there's only DefaultClanTierModel which doesn't seem right. If you guys finds a nice place, please tell me.


TYoung86:
The logic for sandbox shouldn't be in the base dlls, that's a violation of the principles of loosely coupling, in the future they _might_ care enough to move it to the SandBox module.
We'll address it when/if they ever do by doing a version check and `Type.GetType` like in `LordConversationsCampaignBehaviorPatch`.

Issue #28, opened 2020-04-06
Agriculture doesn't work


tynakuh:
Should give 30% more production from farms.
It also applies to 'Personal' when it should probably apply to Governor

PR #75, opened 2020-04-09
Fix for Agriculture

Skau:
Fix for #28

### DIFF
--- a/src/CommunityPatch/Patches/AgriculturePerkPatch.cs
+++ b/src/CommunityPatch/Patches/AgriculturePerkPatch.cs
@@ -1,0 +1,62 @@
+using System;
+using System.Linq;
+using System.Reflection;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.SandBox.GameComponents.Map;
+using TaleWorlds.Core;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches {
+
+  sealed class AgriculturePatch : PatchBase<AgriculturePatch> {
+
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo TargetMethodInfo = typeof(DefaultVillageProductionCalculatorModel).GetMethod("CalculateDailyFoodProductionAmount", BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
+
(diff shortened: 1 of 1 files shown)


src/CommunityPatch/Patches/AgriculturePerkPatch.cs
```diff
@@ -0,0 +1,62 @@
...
+using TaleWorlds.Core;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches {
+
+  sealed class AgriculturePatch : PatchBase<AgriculturePatch> {
+
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo TargetMethodInfo = typeof(DefaultVillageProductionCalculatorModel).GetMethod("CalculateDailyFoodProductionAmount", BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
```
tynakuh, line 16: Should this not be applied on CalculateDailyProductionAmount instead?

src/CommunityPatch/Patches/AgriculturePerkPatch.cs
```diff
@@ -0,0 +1,62 @@
...
+      });
+    }
+
+    // ReSharper disable once InconsistentNaming
+    private static void Postfix(ref float __result, Village village) {
+      var perk = ActivePatch._perk;
+      if (!(village.Bound?.OwnerClan?.Leader?.GetPerkValue(perk) ?? false))
+        return;
+
+      __result *= perk.PrimaryBonus;
```
tynakuh, line 57: I'm assuming result can never be negative here

src/CommunityPatch/Patches/AgriculturePerkPatch.cs
Skau: You might be right. We did go back and forth a bit on which of the two to patch.

src/CommunityPatch/Patches/AgriculturePerkPatch.cs
Skau: In this case, no. Hearth is never negative.

src/CommunityPatch/Patches/AgriculturePerkPatch.cs
tynakuh: > looks like CalculateDailyFoodProductionAmount is actually used to calculate production for food goods in the village in TickFoodProduction in VillageGoodProductionCampaignBehaviour anyways
> so I think CalculateDailyFoodProductionAmount should do the job
> and the result for food production shouldn't be negative, as it just takes the hearthlevel and multiplies it by 0.5 for negative issues

LGTM then

Issue #45, opened 2020-04-06
Agrarian doesn't work


tynakuh:
Should give 30% more production from farms. (inverse of mines?)

PR #77, opened 2020-04-09
Fix Agrarian Patch

zagganoth:
Issue #45


src/CommunityPatch/Patches/StewardAgrarianPatch.cs
```diff
@@ -0,0 +1,90 @@
...
+
+      var hash = bytes.GetSha256();
+      return hash.SequenceEqual(new byte[] {
+        0x08,0x72,0x20,0x9c,0x9a,0x23,
+        0x12,0xdf,0xd7,0xa2,0x83,0x20,
+        0xa5,0x89,0xd7,0x6a,0x67,0xb3,
+        0x35,0x22,0x66,0xb1,0xea,0x2c,
+        0x21,0x00,0x4f,0x45,0x3c,0x74,
+        0x04,0x5e
+      });
```
Tyler-IN, line 49: Could you arrange these in rows of 8?
Makes it easy to count 32 bytes to confirm a full SHA256 hash.
5*6 + 2 is a little weird.

src/CommunityPatch/Patches/StewardAgrarianPatch.cs
```diff
@@ -0,0 +1,90 @@
...
+      if (Applied) return;
+            
+      CommunityPatchSubModule.Harmony.Patch(TargetMethodInfo,
+        postfix: new HarmonyMethod(PatchMethodInfo));
+      Applied = true;
+        
+    }
+    
+    public static void Postfix(ref int __result, Village village) {
+      if (!(village.Bound?.Town?.Governor?.GetPerkValue(PerkObject.FindFirst(x => x.Name.GetID() == "XNc2NIGL")) ?? false)) {
```
Tyler-IN, line 80: Scanning the perk list by using FindFirst has a non-zero performance impact.

I think you had to do this before PatchBase<TPatch> and ActivePatch were introduced.

There are some other patches that use this to work around it.

... Just realised you are getting it like that below... just move that definition up and refer to perk.

Tyler-IN:
Update README.md too

PR #78, opened 2020-04-09
Fixed off-by-one error in default prisoner recruitment model

### LINKED ISSUE #47: Prisoner Recruitment only allows Tier 4 or lower.
Looking at the code:

```
if (!IsPrisonerRecruitable(characterAtIndex) || characterAtIndex.Tier > 6)
{
      continue;
}
```

So Tier 6 shouldn't be recruitable, that's fine but the DefaultPrisonerRecruitmentCalculationModel returns the following:

```
	public override float[] GetDailyRecruitedPrisoners(MobileParty mainParty)
	{
		return new float[5]
		{
			1f,
			0.5f,
			0.3f,
			0.2f,
			0.1f
		};
	}
```
Which is the 'probability' that each tier gets recruited.

However Tier 0 isn't a thing (Well it is, applies to Heroes actually) for troops.
So it's an off by one error, just need to replace that list with something like:

```
	public override float[] GetDailyRecruitedPrisoners(MobileParty mainParty)
	{
		return new float[5]
		{
			1f,
                        1f,
			0.5f,
			0.3f,
			0.2f,
			0.1f
		};
	}

```

for the mod, (Proper fix is of course just decrement the Tier by 1 in the actual code or use a different structure like a Dictionary to determine this, also the method name is a bit confusing...)

tynakuh:
Fixes #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,11 @@ Current Fixes:
 * The Bannerlord perk.
 * The Man-At-Arms perk.
 * The Swords As Tribute (Steward) perk.
+* The Prominence perk.
+* The Supreme Authority perk.
 * Fixes Learning Rate explanation.
 * Fixes Item Comparison perk-based coloring.
+* Fixes Tier 5 Prisoners not being recruit-able. (Note: Tier 6 are not by design)
 
 Current Features:
 * Enable and Disable the Intro Video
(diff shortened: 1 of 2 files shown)


tynakuh:
#79 raised for alternative

Tyler-IN:
Update README.md too

PR #94, opened 2020-04-09
Learning Rate explanation patch IsApplicable check fix

Tyler-IN:
Since this is fixing CalculateLearningRate, not CalculateLearningLimit; there's nothing wrong with it and I'm not changing it's behavior, I need to be checking the hash of CalculateLearningRate instead.

The patches still apply to CalculateLearningLimit

### DIFF
--- a/src/CommunityPatch/Patches/LearningRateExplainerPatch.cs
+++ b/src/CommunityPatch/Patches/LearningRateExplainerPatch.cs
@@ -25,6 +25,14 @@ namespace CommunityPatch.Patches {
       .GetMethod("CalculateLearningLimit", BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly,
         null, CallingConventions.Any, new[] {typeof(int), typeof(int), typeof(TextObject), typeof(StatExplainer)}, null);
 
+    private static readonly MethodInfo CheckMethodInfo1 = typeof(DefaultCharacterDevelopmentModel)
+      .GetMethod("CalculateLearningRate", BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly,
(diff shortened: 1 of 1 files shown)


src/CommunityPatch/Patches/LearningRateExplainerPatch.cs
```diff
@@ -48,15 +56,15 @@ class LearningRateExplainerPatch : IPatch {
...
         var hash = bytes.GetSha256();
         if (!hash.SequenceEqual(new byte[] {
-          0x5B, 0x47, 0x05, 0x79, 0x5A, 0x5D, 0xCB, 0xFC,
-          0xFA, 0x46, 0x21, 0x96, 0x00, 0x94, 0x8D, 0xCC,
-          0xC2, 0xE2, 0x35, 0x7F, 0x70, 0x0D, 0x64, 0xD5,
-          0x08, 0xAA, 0x8F, 0x2B, 0xA5, 0x0B, 0xF3, 0x4B
+          0x74, 0x9e, 0xb0, 0xdf, 0xb1, 0x47, 0x44, 0x6a,
+          0xb9, 0x95, 0x50, 0xef, 0x11, 0x36, 0xdf, 0x2f,
+          0xd9, 0x86, 0x52, 0xf6, 0xa8, 0x18, 0x7b, 0x33,
+          0x8f, 0x5c, 0x51, 0xd8, 0x16, 0xa5, 0xe8, 0xd6
```
zagganoth, line 67: should we include the old hashes to ensure compatibility with older versions? We're doing this for some of the other patches I think

Skau: I think that we should just keep whatever hash is currently there and just keep on stacking them when the IL changes. Commenting above what version the hash is for is probably nice to have to.

Skau: And the hash is changed instead of added here because he changed what method to check the hash for.

PR #96, opened 2020-04-09
Report Generation for Exceptions in Tick Finalizer

Tyler-IN:
Similar but wider reaching than Bannerlord Tweaks' version.

Since it's using a finalizer, it's (almost) guaranteed to run if an exception occurs somewhere beneath the call frame.

Considering any possible ways to hook into a native crash handler too.


tynakuh:
Looks great, I've been doing something similar in my other mod, should've thought to do it here!

Issue #35, opened 2020-04-06
Prosperous Reign doesn't work


tynakuh:
Villages don't 'grow' at double rate.  Question of what is growth  exactly.  I'd imagine it's Hearth per daily tick.

PR #103, opened 2020-04-10
Prosperous Reign patch

Skau:
Fix for #35

### DIFF
--- a/src/CommunityPatch/Patches/ProsperousReignPatch.cs
+++ b/src/CommunityPatch/Patches/ProsperousReignPatch.cs
@@ -62,7 +62,7 @@ namespace CommunityPatch.Patches {
 
       var explainedNumber = new ExplainedNumber(__result, explanation);
       
-      if (explanation.Lines.Count > 0)
+      if (!(__result > -0.0001f) || !(__result < 0.0001f) && explanation.Lines.Count > 0)
         explanation.Lines.RemoveAt(explanation.Lines.Count - 1);
       
       float extra = explanation.Lines.Where(line => line.Number > 0).Sum(line => line.Number);
--- a/src/CommunityPatch/Patches/ProsperousReignPatch.cs
+++ b/src/CommunityPatch/Patches/ProsperousReignPatch.cs
@@ -1,0 +1,79 @@
+using System;
+using System.Linq;
+using System.Reflection;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.SandBox.GameComponents;
+using TaleWorlds.Core;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches {
+
+  sealed class ProsperousReignPatch : PatchBase<ProsperousReignPatch> {
+
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo TargetMethodInfo = typeof(DefaultSettlementProsperityModel).GetMethod("CalculateHearthChangeInternal", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly);
+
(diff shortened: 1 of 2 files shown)


src/CommunityPatch/Patches/ProsperousReignPatch.cs
```diff
@@ -0,0 +1,79 @@
...
+        // e.1.0.9
+        0x65, 0x10, 0x58, 0xD9, 0xC7, 0xFA, 0x99, 0x07, 
+        0xB5, 0x92, 0xC0, 0xB9, 0xC4, 0xEF, 0x94, 0xBF, 
+        0x5D, 0x7B, 0xCA, 0xA0, 0x1E, 0x37, 0xB9, 0x73, 
+        0xFA, 0x2C, 0x1D, 0x91, 0xC0, 0x5F, 0xF1, 0x14
+      });
+    }
+
+    // ReSharper disable once UnusedParameter.Local
+    private static void Prefix(Village village, ref StatExplainer explanation)
```
tynakuh, line 54: Do prefixes not need to return a bool?

src/CommunityPatch/Patches/ProsperousReignPatch.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    // ReSharper disable once InconsistentNaming
+    private static void Postfix(ref float __result, Village village, StatExplainer explanation) {
+      var perk = ActivePatch._perk;
+      if (!(village.Bound?.OwnerClan?.Leader?.GetPerkValue(perk) ?? false))
+        return;
+
+      var explainedNumber = new ExplainedNumber(__result, explanation);
+      
+      if (explanation.Lines.Count > 0)
```
tynakuh, line 65: Should we also check if __result was 0, since that won't add a description to explaination if I recall correctly
(If it's something like +1, -1 we'd delete the -1)

src/CommunityPatch/Patches/ProsperousReignPatch.cs
Skau: No, a void prefix does the same as returning true

tynakuh: Ah okay, cool.

src/CommunityPatch/Patches/ProsperousReignPatch.cs
Skau: You mean an early return if __result is 0?

tynakuh: No, new ExplainedNumber(__result, explanation) won't add an additional description if its 0:

```	if (!(value > -0.0001f) || !(value < 0.0001f))
	{
		Explainer.AddLine((baseExplanier != null) ? baseExplanier.ToString() : baseStr.ToString(), value, StatExplainer.OperationType.Base);
	}```

Skau: I'm not following :/

tynakuh: Looking at implementation of the original method don't think its possible anyways but i'll keep going haha.

       `explanation.Lines.RemoveAt(explanation.Lines.Count - 1);`

^ That line could get executed despite no 'Base +n' getting added

tynakuh: Oh wait maybe i'm being dumb.

Skau: I see now. So `if (__result != 0 && explantion.Lines.Count > 0)`?

tynakuh: Yea, though `(!(value > -0.0001f) || !(value < 0.0001f))` if we want to follow their logic to not add a line

Tyler-IN: Check was added, resolving this convo.

Issue #34, opened 2020-04-06
Logistics Expert doesn't work


tynakuh:
Armies do not gain speed from this perk. The amount is ambigious so should probably be configurable or discussed.

PR #104, opened 2020-04-10
Fixed Logistics Expert

Skau:
Fixes issue #34 
Even though the text doesn't mention the increase in speed, the perk is initialized with a en effect of 0.1, so I just went with that. We can add an option for it if it's wanted.

### DIFF
--- a/src/CommunityPatch/Patches/LogisticsExpertPatch.cs
+++ b/src/CommunityPatch/Patches/LogisticsExpertPatch.cs
@@ -36,12 +36,12 @@ namespace CommunityPatch.Patches {
 
       var bytes = TargetMethodInfo.GetCilBytes();
       if (bytes == null) return false;
-      
+
       var hash = bytes.GetSha256();
       return hash.SequenceEqual(new byte[] {
-        0x06, 0xE6, 0xB4, 0xA5, 0xA5, 0x0C, 0xAD, 0x4E, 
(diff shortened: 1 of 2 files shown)


src/CommunityPatch/Patches/LogisticsExpertPatch.cs
```diff
@@ -0,0 +1,64 @@
...
+        0x06, 0xE6, 0xB4, 0xA5, 0xA5, 0x0C, 0xAD, 0x4E,
+        0x5C, 0xCB, 0xB8, 0xBD, 0x80, 0x43, 0x52, 0x81,
+        0x70, 0x70, 0x34, 0x4E, 0x14, 0xA0, 0x58, 0x93,
+        0xAB, 0x0C, 0x84, 0xAC, 0x43, 0x3F, 0xDD, 0x63
+      });
+    }
+
+    // ReSharper disable once UnusedParameter.Local
+    // ReSharper disable once InconsistentNaming
+    private static void Postfix(ref float __result, MobileParty mobileParty, float baseSpeed, StatExplainer explanation) {
```
tynakuh, line 51: Could technically drop the unused 'baseSpeed'

Skau: Ah right. I wasn't aware you could drop unused parameters. I just assumed they had to be there because if the naming is off it doesn't work.

PR #113, opened 2020-04-10
Patches for Mounted Archer, Bow Expert, Crossbow Cavalry, and Crossbow Expert

ajseavey:
Addresses #10 

Heavily based on my crossbow cavalry perk, mostly just a change in condition and the field being changed. Currently the only difference in the item usage between bows and long bows is whether they are able to be used on mounts. This just will conditionally overwrite the item usage of a long bow with that of a bow based on the Agent's perks.


src/CommunityPatch/Patches/AgentWeaponEquippedPatch.cs
```diff
@@ -0,0 +1,68 @@
...
+      if (Applied) return false;
+
+      var patchInfo = Harmony.GetPatchInfo(TargetMethodInfo);
+      if (AlreadyPatchedByOthers(patchInfo))
+        return false;
+
+      return _IsApplicable(game);
+    }
+
+    private static void PostFix(ref Agent __instance,
```
Tyler-IN, line 38: Given the Patch call above now applies this as the prefix parameter, does this need to be renamed to Prefix?

Also, postfix (proper; suffix) and prefix are full words, so Pascal naming convention would be to just capitalize the P.

src/CommunityPatch/ApplicationVersionComparer.cs
```diff
@@ -25,8 +25,10 @@ public class ApplicationVersionComparer : IComparer<ApplicationVersion>, IEquali
     public bool Equals(ApplicationVersion x, ApplicationVersion y)
       => Compare(x, y) == 0;
 
+    public bool GreaterThan(ApplicationVersion x, ApplicationVersion y) => Compare(x, y) > 0;
+
     public int GetHashCode(ApplicationVersion obj)
-      => ((int) _typeComparer.GetHashCode(obj.ApplicationVersionType) << 28) | (obj.Major << 23) | (obj.Minor << 12) | obj.Revision;
+    => ((int) _typeComparer.GetHashCode(obj.ApplicationVersionType) << 28) | (obj.Major << 23) | (obj.Minor << 12) | obj.Revision;
 
   }
```
Tyler-IN, line 33: Can you revert these minor formatting changes, they might be accidentally committed?
They're out of scope for the PR.

Tyler-IN:
Title of PR needs to be adjusted to account for the merge for record keeping.

Other issues need to be tagged.

src/CommunityPatch/Patches/AgentWeaponEquippedPatch.cs
ajseavey: Fixed

src/CommunityPatch/ApplicationVersionComparer.cs
ajseavey: Fixed

Tyler-IN: This shouldn't show up in the PR as modified...

PR #122, opened 2020-04-11
Fixed Swords and Tribute to only be applied to rulers

Skau:
Tested it and looks good now. Fixed because a comment on nexusmods mentioned that it applied even when the player was just a vassal.


Skau:
Right, I just semi forgot about it. I just wanted another opinion before pushing.

Issue #46, opened 2020-04-06
Reconstruction doesn't work


tynakuh:
Villages should recover 'faster' than normal after being raided.

The actual number isn't stated, so should probably be configurable or discussed.

PR #128, opened 2020-04-13
Reconstruction Perk Fix

Skau:
Fix for #46 
The amount was ambigious. Neither the description or the perk initialization had a value. The default value passed in for healing is hardcoded in `VillageHealCampaignBehavior.DailyTick` (0.06), so I ended up multiplying this by 2. 
The value could probably be bigger though, this is a 275 skill perk.
I am currently on e1.0.10, so that's the only hash, if anyone is on the beta branch and wants to quickly check if the hash is different, please do.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Steward/ReconstructionPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Steward/ReconstructionPatch.cs
@@ -1,0 +1,92 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using Actions;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using static CommunityPatch.HarmonyHelpers;
+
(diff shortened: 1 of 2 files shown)


src/CommunityPatch/Patches/Perks/Intelligence/Steward/ReconstructionPatch.cs
```diff
@@ -0,0 +1,92 @@
...
+      if (AlreadyPatchedByOthers(patchInfo))
+        return false;
+
+      var hash = TargetMethodInfo.MakeCilSignatureSha256();
+      return hash.MatchesAnySha256(Hashes);
+    }
+
+    // ReSharper disable once InconsistentNaming
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void Prefix(Settlement settlement, ref float percentage) {
```
zagganoth, line 80: maybe this should be renamed to postfix to avoid confusion

Skau: Thanks. It is a prefix. I just had a rearrangement just before I pushed, so it was pushed as a postfix.

Issue #43, opened 2020-04-06
Food Rationing doesn't work


tynakuh:
If you are under siege, food consumption in settlement should be reduced by 40%.

PR #132, opened 2020-04-13
Fixed Food Rationing Perk

Skau:
Fix for #43


src/CommunityPatch/Patches/Perks/Intelligence/Steward/FoodRationingPatch.cs
```diff
@@ -0,0 +1,82 @@
...
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void Prefix(ref StatExplainer tooltipStringBuilder)
+      => tooltipStringBuilder ??= new StatExplainer();
+
+    // ReSharper disable once InconsistentNaming
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void Postfix(ref float __result, Town town, StatExplainer tooltipStringBuilder) {
+      var perk = ActivePatch._perk;
+      var leader = town.Settlement?.OwnerClan?.Leader;
+      var settlement = town.Owner?.Settlement;
```
zagganoth, line 69: Quite likely I'm missing something - but wouldn't it be simpler just to check town.IsUnderSiege and avoid the reference to settlement entirely?

Issue #87, opened 2020-04-09
Extra Arrows / Throwing Weapons perks not working


ExoLyzix:
Those perks aren't working : 

- Bow tree : Large Quiver, Battle Equipped

- Throwing : Fully Armed, Battle Ready, Well Prepared ( Not sure about that one )

- Riding : Spare Arrows, Spare Throwing Weapons

- Athletics : Extra Arrows, Extra Throwing Weapons

They're all supposed to give you a certain amount of arrows / throwing weapons but they don't, I didn't see any issues posted about it

ajseavey:
I don't know if party members even have projectile limits enforced. I'll be taking a look at this in the near future.

fnzr:
@ajseavey I've been looking into this, so if you haven't started working on it yet, I think I got it.

ajseavey:
Which part? There are changes coming soon that may make this easier.

fnzr:
I'm looking at bow first, but all of extra ammo perks should follow pretty much the same logic.
I'm using Harmony to patch `Agent::InitializeMissionEquipment`

ajseavey:
Check the current master. Stuff related to Agent::WeaponEquipped just got merged that could also make this a lot simpler.

fnzr:
Alright, I'll make sure to check it.

fnzr:
This should be closed and then opened a issue exclusively for the Well Prepared perk in the Throwing tree.

PR #135, opened 2020-04-14
Extra ammo perk fixes

fnzr:
Fixes the following perks mentioned in #87:
* Bow
    Large Quiver
    Battle Equipped

* Throwing
 Fully Armed
 Battle Ready

* Riding
 Spare Throwing Weapon
 Spare Arrows

* Athletics
 Extra Arrows
 Extra Throwing Weapon

Since this patch covers several skill trees, I'm not sure where to put it, so it might be necessary to move it before merging.
The issue also referenced the perk Well Prepared on the Throwing tree, but that perk, if in fact isn't working, will require a different approach and I'm not sure if it's even worth to fix it with a patch instead of letting TW handle it, but this can be discussed later.


TYoung86:
TaleWorlds could possibly have tickets for each perk, and only fix a portion instead of the whole thing.

Please break this patch up per perk. Otherwise, it looks good.

fnzr:
I made this a single patch instead of multiple small ones is because shared code, yes, but also performance considerations. `MissionWeapon` is a struct, so each time I modify the value of the object I have to create a whole copy of the object.
The way it's done, at most one copy is created and assigned. Doing it per perk increases this value up to 5.
Also I use Postfixes instead of prefixes to guarantee that I'm modifying the correct agent.Equipment without extra checks, but it doesn't make much of a difference either way.

Tyler-IN:
If MissionWeapon is a struct, use a ref to pass it around so it doesn't get copied?

src/CommunityPatch/Patches/ExtraAmmoPerksPatch.cs
```diff
@@ -0,0 +1,147 @@
...
+        var missionWeapons = (MissionWeapon[]) property.GetValue(__instance.Equipment);
+
+        //At this point agent.HasMount wasn't initialized, and trying to modify ammo in later functions causes problems
+        var item = __instance.SpawnEquipment[EquipmentIndex.ArmorItemEndSlot].Item;
+        var hasMount = item != null && item.HasHorseComponent;
+
+        var hero = charObj.HeroObject;
+        for (var i = 0; i < missionWeapons.Length; i++) {
+          if (!missionWeapons[i].Weapons.IsEmpty()) {
+            var weaponComponentData = missionWeapons[i].Weapons[0];
```
Tyler-IN, line 119:
```suggestion
            ref var weaponComponentData = ref missionWeapons[i].Weapons[0];
```

src/CommunityPatch/Patches/ExtraAmmoPerksPatch.cs
```diff
@@ -0,0 +1,147 @@
...
+              else if (weaponComponentData.WeaponClass == WeaponClass.ThrowingAxe ||
+                weaponComponentData.WeaponClass == WeaponClass.ThrowingKnife ||
+                weaponComponentData.WeaponClass == WeaponClass.Javelin) {
+                extraAmmo = ApplyThrowingAmmoPerks(hero, hasMount);
+              }
+
+              if (extraAmmo > 0) {
+                var maxAmmoField = typeof(MissionWeapon).GetField("_maxDataValue", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.DeclaredOnly);
+                var ammoField = typeof(MissionWeapon).GetField("_dataValue", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.DeclaredOnly);
+                var newMaxAmmo = (short) ((short) maxAmmoField.GetValue(missionWeapons[i]) + extraAmmo);
```
Tyler-IN, line 135: We'd probably need a get ref of field helper here.

fnzr:
> This doesn't seem to be invoked every tick.
This is invoked only once, when starting the mission.

>what performance impact has been measured?
Fair enough, it just felt unnecessary when writing the code. I'll make the proposed changes and report back.

src/CommunityPatch/Patches/ExtraAmmoPerksPatch.cs
Tyler-IN: (If performance is a concern.)

src/CommunityPatch/Patches/ExtraAmmoPerksPatch.cs
Tyler-IN: (If performance is a concern.)

Tyler-IN:
The `ref` proposed changes aren't necessary if performance isn't a concern, that splitting up bit though. :)

src/CommunityPatch/Patches/ExtraAmmoPerksPatch.cs
```diff
@@ -0,0 +1,122 @@
...
+      var hash = TargetMethodInfo.MakeCilSignatureSha256();
+      return hash.MatchesAnySha256(Hashes);
+    }
+
+    protected static bool HasMount(Agent agent) {
+      var item = agent.SpawnEquipment[EquipmentIndex.ArmorItemEndSlot].Item;
+      return item != null && item.HasHorseComponent;
+    }
+
+    static short ApplyArrowPerks(Hero hero, bool hasMount) {
```
ajseavey, line 50: This and the other similar function seem unnecessary now.

src/CommunityPatch/Patches/Perks/Control/Throwing/FullyArmedPatch.cs
```diff
@@ -0,0 +1,28 @@
...
+    public override void Apply(Game game) {
+      if (Applied) return;
+
+      CommunityPatchSubModule.Harmony.Patch(TargetMethodInfo,
+        postfix: new HarmonyMethod(PatchMethodInfo));
+      Applied = true;
+    }
+
+    static bool CanApplyPerk(Hero hero, WeaponComponentData weaponComponentData)
+      => WeaponComponentData.GetItemTypeFromWeaponClass(weaponComponentData.WeaponClass) == ItemObject.ItemTypeEnum.Thrown &&
```
ajseavey, line 22: Nit: Move the arrow up a line to format it similar to the other patches

src/CommunityPatch/Patches/ExtraAmmoPerksPatch.cs
fnzr: yup, I'm still working on it.

src/CommunityPatch/Patches/Perks/Control/Bow/BattleEquippedPatch.cs
```diff
@@ -0,0 +1,28 @@
...
+      CommunityPatchSubModule.Harmony.Patch(TargetMethodInfo,
+        postfix: new HarmonyMethod(PatchMethodInfo));
+      Applied = true;
+    }
+
+    static bool CanApplyPerk(Hero hero, WeaponComponentData weaponComponentData) =>
+      WeaponComponentData.GetItemTypeFromWeaponClass(weaponComponentData.WeaponClass) == ItemObject.ItemTypeEnum.Arrows &&
+      hero.GetPerkValue(DefaultPerks.Bow.BattleEquipped);
+    
+    private static void Postfix(Agent __instance) => ApplyPerk(__instance, 6, CanApplyPerk);
```
ajseavey, line 25: Also I think this and the other `Postfix` methods could use the `[MethodImpl(MethodImplOptions.NoInlining)]` we have elsewhere. You may or may not already be working on that, and if you are feel free to ignore this.

fnzr:
Should be good to go now.

PR #150, opened 2020-04-15
Fix user not being told that FirstPhase story quests fail in 10 years

### LINKED ISSUE #133: Tell user that early story quests will fail after 10 years
Suggestion is to include somewhere in the UI telling the user that the first phase story line quests (everything up to and including Create / Support Kingdom) will fail after 10 years (which I guess in bannerlord time is 840 days ish). Judging by many posts on taleworlds forums and reddit people are getting blindsided by this.

P.S. Still true after patch 1.0.11, if anyone's wondering.

iPherian:
Fixes #133 . Now those quests will appear with the days until failure much like other quests.

Unfortunately there's a few hours game time ( 10 seconds or so real life time) until the quests are updated after getting them. See code [comments](https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/pull/150/commits/0cc24f958e1d111d549891ffe08af49e0f40ae94#diff-4cb347780d4b8f8827fabdb6f694148dR105).

### DIFF
--- a/src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
+++ b/src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
@@ -1,0 +1,113 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using StoryMode;
+using StoryMode.Behaviors;
+using StoryMode.StoryModePhases;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+
+namespace CommunityPatch.Patches {
+
+  public sealed class EarlyStoryVisibleTimeoutPatch : IPatch {
+    // default if we can't obtain vanilla's value
+    private const int DefaultFirstPhaseTimeLimitInYears = 10;
+    private readonly int FirstPhaseTimeLimitInYears;
+
+    public EarlyStoryVisibleTimeoutPatch() {
+      FirstPhaseTimeLimitInYears = ExtractFirstPhaseTimeLimitInYears();
+    }
+
+    public bool Applied { get; private set; }
+
+    public void Reset() {
+      if (Applied) {
+        CampaignEvents.OnQuestStartedEvent.ClearListeners(this);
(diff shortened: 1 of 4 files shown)


Tyler-IN:
Should refactor this into a separate IPatch inheriting class similar to DisciplinarianPerkPatch fix that doesn't use Harmony. Need to find where the time frame is in code and extract it instead of hard-coding FirstPhaseTimeLimitInYears.

iPherian:
Go ahead.

iPherian:
Thought i'd take a stab at it. Moved patch to class EarlyStoryVisibleTimeoutPatch. Can squash the commits if necessary.

iPherian:
@Tyler-IN  Is that good, do you think?

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
```diff
@@ -0,0 +1,113 @@
...
+
+      // set visible timeout to be when vanilla would have (silently) timed 
+      // out the quest,  minus a day to make very sure the quest doesn't 
+      // somehow trigger vanilla's silent timeout too early.
+      CampaignTime newDueTime = FirstPhase.Instance.FirstPhaseStartTime
+                                  + CampaignTime.Years(FirstPhaseTimeLimitInYears)
+                                  - CampaignTime.Days(1);
+      if (quest.QuestDueTime != newDueTime) {
+        quest.ChangeQuestDueTime(newDueTime);
+        ShowNotification(new TextObject("{=!}Quest time remaining was updated."), "event:/ui/notification/quest_update");
```
Tyler-IN, line 80:
```suggestion
        ShowNotification(new TextObject("{=QuestTimeRemainingUpdated}Quest time remaining was updated."), "event:/ui/notification/quest_update");
```

Tyler-IN: even if there's no localization for it, we can add a localization file for it later

iPherian:
@Tyler-IN should I change the PR to be merged into dev branch, or keep it going into master?

PR #164, opened 2020-04-15
Base removed + fixed Bannerlord and ManAtArms

### LINKED ISSUE #163: Man-at-Arms and Bannerlord are calculating extra? - Perks  - Steward
Steward perks Man-at-Arms & Bannerlord, fixed in an earlier patch seems to be calculating the numbers a bit excessive?
As per the Kingdom page the fiefs are castles and cities only while towns are not considered fiefs, but the perk considers them as fiefs as well, is this intentional or just a bug?

Also a minor graphical issue, Base amount appears twice in the list, once at the top, then a sum of above values before the perks, however they dont double count, but it just appears.

game version 1.0.11

image

Skau:
Resolves #163.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Endurance/Riding/ConroiPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Endurance/Riding/ConroiPatch.cs
@@ -63,10 +63,14 @@ namespace CommunityPatch.Patches.Perks.Endurance.Riding {
       var perk = ActivePatch._perk;
 
       var extra = perk.PrimaryBonus * party.MemberRoster.Count(x => x.Character.IsHero && x.Character.HeroObject.GetPerkValue(perk));
-      if (extra <= 0) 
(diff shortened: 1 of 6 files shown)


Tyler-IN:
```cs


		// Token: 0x17000835 RID: 2101
		// (get) Token: 0x06001ADE RID: 6878 RVA: 0x00069815 File Offset: 0x00067A15
		public override bool IsTown
		{
			get
			{
				return !this._isCastle;
			}
		}

		// Token: 0x17000836 RID: 2102
		// (get) Token: 0x06001ADF RID: 6879 RVA: 0x00069820 File Offset: 0x00067A20
		public override bool IsCastle
		{
			get
			{
				return this._isCastle;
			}
		}
```

IsCastle works, IsTown does not.

Skau:
Swapped it around and tested for !IsVillage instead, and after testing it works as expected.

PR #167, opened 2020-04-16
Add implementation to the Construction Expert perk

miguelcjalmeida:
I have added the engineering construction expert missing perk. Also, updated the readme.md with new perk category named 'Engineering'. 

The implementation reduces the production cost (in 30%) for walls and fortifications when the governor of the castle/city has the construction expert perk active. It also takes in consideration that the 'castle charters' policy might be active, which reduces the cost in 20% by itself. Having both the perk and policy active, will plummet the costs by 50% additively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ This is going to be a mod that just fixes up some things in Mount &amp; Blade 2:
     * Tax Collector
     * Warmonger
     * War Rations
+  * Engineering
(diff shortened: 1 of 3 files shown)


miguelcjalmeida:
I just realized the perk should be increasing the production speed not the cost. Gonna fix that and close this pull request

Issue #141, opened 2020-04-14
A hidden perk beneath the Legendary Smith perk


buggiant:
There‘s a hidden perk beneath the Legendary Smith perk.
It shines there but is unable to click, and the character menu keeps reminding me I have one perk to upgrade, it really bothers me.

wrobell89:
As a quick fix for, you can select this perk with Character Export mod. Just export the character, mark perk as selected and import.

PR #172, opened 2020-04-17
Fix hidden Smithing Perk

fnzr:
Solves #141.
Siege Expert


tynakuh:
Code LGTM but Siege Expert perk itself does nothing I believe.

fnzr:
Updated. I'll see if I can fix the Siege Expert perk because having a note saying it doesn't work feels bad.

Issue #120, opened 2020-04-11
Talk to other lords when encountering armies


dijicaek:
Currently it seems you can only talk to other lords in an army if you're part of their faction.

I'm a complete novice when it comes to programming, let alone modding, otherwise I'd attempt to fix it myself but I did find "game_menu_army_talk_to_other_members" within EncounterGameMenuBehavior.cs.

Perhaps this would be a starting point for someone to go off if they wanted to fix it? That is, assuming it is unintended behaviour and within the scope of this mod to fix.

McEagle:
I have been looking for a way to fix this.

## Use cases

The player will basically get a menu when interacting with an army of its faction and a conversation if the player isn't part of it (enemy or neutral).

Therefore, without the menu, the player will directly talk to the army's leader without having the option to talk to the other members.

## Possible solution

So a quick fix is to allow this menu to pop up for each faction when encountering an army. Taleworlds implemented the menu in a way that the player can talk to other members (if any) of a friendly army. An attack/leave option is available when encountering a enemy army.

The ```DoMeetingInternal``` method in the ```TaleWorlds.CampaignSystem.PlayerEncounter``` class is responsible for this behaviour : 
```csharp
private void DoMeetingInternal()
{
  PartyBase party1 = this._encounteredParty;
  if (party1.IsSettlement)
  {
    foreach (PartyBase party2 in (IEnumerable<PartyBase>) MobileParty.MainParty.MapEvent.DefenderSide.Parties)
    {
      if (!party2.IsSettlement)
      {
        party1 = party2;
        break;
      }
    }
  }
  Campaign.Current.CurrentConversationContext = ConversationContext.PartyEncounter;
  this.EncounterState = PlayerEncounterState.Begin;
  this._stateHandled = true;
  if (PlayerEncounter.PlayerIsAttacker 
    && this._defenderParty.IsMobile
    && (this._defenderParty.MobileParty.Army != null
      && this._defenderParty.MobileParty.Army.LeaderParty == this._defenderParty.MobileParty)
    && (this._defenderParty.MobileParty.Army.LeaderParty.MapFaction == MobileParty.MainParty.MapFaction
      && !this._defenderParty.MobileParty.Army.LeaderParty.AttachedParties.Contains(MobileParty.MainParty)))
  {
    GameMenu.SwitchToMenu("army_encounter");
  }
  else
  {
    this._meetingDone = true;
[...]

PR #177, opened 2020-04-17
Menu when encountering an army

JoeFwd:
Implementation of https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/issues/120#issuecomment-613501060

A menu will pop up when encoutering a **neutral** army enabling the player to : 
* talk to the army's leader
* talk to the other members
* leave

A menu will pop up when encoutering an **enemy** army enabling the player to : 
* talk to the army's leader
* attack
* leave

### DIFF
--- a/src/CommunityPatch/Patches/MenuWhenEncountingAnArmyPatch.cs
+++ b/src/CommunityPatch/Patches/MenuWhenEncountingAnArmyPatch.cs
@@ -72,14 +72,14 @@ namespace CommunityPatch.Patches {
         return true;
       }
       
-      PartyBase party1 = (PartyBase) EncounteredPartyField.GetValue(__instance);
-      if (party1.IsSettlement)
+      PartyBase attacker = (PartyBase) EncounteredPartyField.GetValue(__instance);
+      if (attacker.IsSettlement)
       {
-        foreach (PartyBase party2 in (IEnumerable<PartyBase>) MobileParty.MainParty.MapEvent.DefenderSide.Parties)
+        foreach (PartyBase defender in (IEnumerable<PartyBase>) MobileParty.MainParty.MapEvent.DefenderSide.Parties)
         {
-          if (!party2.IsSettlement)
+          if (!defender.IsSettlement)
           {
-            party1 = party2;
+            attacker = defender;
             break;
           }
         }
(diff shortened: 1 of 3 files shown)


Tyler-IN:
I approved the code, but does this need to be an option-toggled feature or is it TW's intention?

src/CommunityPatch/Patches/MenuWhenEncountingAnArmyPatch.cs
```diff
@@ -0,0 +1,105 @@
...
+    
+    
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static bool Prefix(PlayerEncounter __instance) {
+      if (EncounteredPartyField == null || MapEventStateField == null || StateHandledField == null || DefenderPartyField == null || MeetingDoneField == null) {
+        CommunityPatchSubModule.Error($"{typeof(MenuWhenEncounteringAnArmyPatch).Name}: Could not locate all of necessary private fields for patching." + Environment.NewLine);
+        return true;
+      }
+      
+      PartyBase party1 = (PartyBase) EncounteredPartyField.GetValue(__instance);
```
tynakuh, line 75: party1 vs party2 aren't very descriptive and can make it a bit harder to follow imo.

JoeFwd: Tbh I left the vanilla implementation names.

Tyler-IN: We can refactor the naming to attacker/defender or whatever this after it gets merged to dev, or...

src/CommunityPatch/Patches/MenuWhenEncountingAnArmyPatch.cs
```diff
@@ -0,0 +1,105 @@
...
+      if (party1.IsSettlement)
+      {
+        foreach (PartyBase party2 in (IEnumerable<PartyBase>) MobileParty.MainParty.MapEvent.DefenderSide.Parties)
+        {
+          if (!party2.IsSettlement)
+          {
+            party1 = party2;
```
Tyler-IN, line 82:
```suggestion
      PartyBase attacker = (PartyBase) EncounteredPartyField.GetValue(__instance);
      if (attacker.IsSettlement)
      {
        foreach (PartyBase otherParty in (IEnumerable<PartyBase>) MobileParty.MainParty.MapEvent.DefenderSide.Parties)
        {
          if (!otherParty.IsSettlement)
          {
            attacker = otherParty;
```

Tyler-IN:
Holding this PR for now, this should probably be an opt-in feature, right?

Tyler-IN:
Opt-in means it should be disabled by default, `DisableMenuWhenEncouteringAnArmy ` defaults to false right? So you should probably flip all the checks and rename it to `EnableMenuWhenEncouteringAnArmy` ... At current, it looks like it's opt-out instead of opt-in.

JoeFwd:
> 
> 
> Opt-in means it should be disabled by default, `DisableMenuWhenEncouteringAnArmy ` defaults to false right? So you should probably flip all the checks and rename it to `EnableMenuWhenEncouteringAnArmy` ... At current, it looks like it's opt-out instead of opt-in.

Yes, you are right. I'll flip the opt !

PR #178, opened 2020-04-17
add implementation for Ballistics perk from Engineering tree

miguelcjalmeida:
I have patched two different methods: SandBoxUIHelper.GetSiegeEngineTooltip and SiegeEvent.BombardHitEngine. 

The first is responsible for creating the siege units tooltips. I added a new tooltip property saying when the Ballistic perk is active. It only shows up when some leader from the defender team have the perk activated, and player is previewing the Catapult. It takes into consideration that catapults are only placeable in the defensor team. 

The latter method is the one that actually applies the damage. Before applying the damage, I make a calculation to first apply a bonus damage. This extra damage only occurs when the conditions are match: some leader hero from the catapult team must have the Ballistic Perk activated.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
@@ -89,7 +89,7 @@ namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
         _perk.Skill,
         (int) _perk.RequiredSkillValue,
         _perk.AlternativePerk,
-        _perk.PrimaryRole, _perk.PrimaryBonus,
+        SkillEffect.PerkRole.PartyLeader, _perk.PrimaryBonus,
         _perk.SecondaryRole, _perk.SecondaryBonus,
         _perk.IncrementType
       );
@@ -126,7 +126,7 @@ namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
       var perk = ActivePatch._perk;
 
       if (!IsCatapult(attackerEngineType)) return;
-      if (!DefendersHaveBallisticPerk(siegeEventSide.SiegeParties)) return;
+      if (!PartiesHaveBallisticPerk(siegeEventSide.SiegeParties)) return;
       
       var bonusDamage = attackerEngineType.Damage * perk.PrimaryBonus;
       damagedEngine.SetHitpoints(damagedEngine.Hitpoints - bonusDamage);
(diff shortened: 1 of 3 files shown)


tynakuh:
Could we discuss this further? While Defenders have explicitly 'catapults' I think the intention behind this is that all catapult type siege equipment will gain 30% damage. 

I have two reasons for this really:
1. Onagers/Trebuchets are all types of catapults.
2. The perk picture itself shows trebuchets
image

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
```diff
@@ -0,0 +1,159 @@
...
+
+    private static bool IsCatapult(SiegeEngineType engine)
+      => engine == DefaultSiegeEngineTypes.Catapult || engine == DefaultSiegeEngineTypes.FireCatapult;
+    
+    private static bool DefendersFromHeroSiegeHaveBallisticPerk() {
+      var perk = ActivePatch._perk;
+      
+      if (Hero.MainHero == null) return false;
+      
+      if (Hero.MainHero.PartyBelongedTo.SiegeEvent == null || Hero.MainHero.PartyBelongedTo.Party.Side == BattleSideEnum.Defender)
```
tynakuh, line 143: Should we not be checking for the party's effective Engineer instead of just the leader hero?

Skau:
This looks hardcoded for only being applied to the main hero. Shouldn't any leader that have this perk get the bonus, and not just the player?

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
Skau: According to the initialization the perk has the personal role. That might be wrong. Should probably be party leader or engineer.

miguelcjalmeida: Responding you tynakuh, that will depend on the perk role. For now it is 'Personal'. Should I change it?

miguelcjalmeida:
> This looks hardcoded for only being applied to the main hero. Shouldn't any leader that have this perk get the bonus, and not just the player?

The hardcode is just for the tooltip displayment, the damage calculations are working for the other leaders too.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
tynakuh: Ah right, hmm. That's a strange perk role, not sure what the best approach is here, Engineer makes the most logical sense to me, but party leader is valid as well. Personal makes no real sense to me (as it doesn't apply to the 'person' specifically...unless we're talking in combat usage of the catapult?)

tynakuh: I'm happy for this to go in as Party Leader wihtout logic change, so long as we change the perk description/property alongside to make it clear to players.

miguelcjalmeida: I have changed the role to Party Leader as suggested. Also I fixed the checking for party leader in the defensor team for displaying the Tooltip. In this case, when the player was in the defensor team, the code was only checking if the hero had the perk active. Now it checks every defender party before displaying the tooltip.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
```diff
@@ -0,0 +1,156 @@
...
+    };
+
+    public override void Reset()
+      => _perk = PerkObject.FindFirst(x => x.Name.GetID() == "LyVZYGkN");
+
+    public override bool? IsApplicable(Game game)
+      // ReSharper disable once CompareOfFloatsByEqualityOperator
+    {
+      if (_perk == null) return false;
+      if (_perk.PrimaryBonus != 0.3f) return false;
```
Tyler-IN, line 63: I missed this. Why is this checked?

PR #179, opened 2020-04-17
Add perk 'Construction Expert' of Engineering tree

miguelcjalmeida:
I am adding the implementation for the 'Construction Expert' perk, which should increase production speed by 30%.

In order to do that, it was necessary to patch two methods from DefaultBuildingConstructionModel. Both patches grab the final result of a calculation and add those extra 30%.

Also, I have added a fix for the project 'GeneratePatchTargetHashes', which was throwing an error on executing in my machine.


src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ConstructionExpertPatch.cs
```diff
@@ -0,0 +1,122 @@
...
+
+    public override void Apply(Game game) {
+      var textObjStrings = TextObject.ConvertToStringList(
+        new List<TextObject> {
+          _perk.Name,
+          _perk.Description
+        }
+      );
+      // most of the properties of skills have private setters, yet Initialize is public
+      _perk.Initialize(
```
tynakuh, line 77: What's the reason for re-initializing the perk here? Maybe i'm not spotting it but seems to re-initialize identically?

miguelcjalmeida: The initialize is indeed identical. There were no need for reinitializing it. I'll have that removed

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ConstructionExpertPatch.cs
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    public override void Reset()
+      => _perk = PerkObject.FindFirst(x => x.Name.GetID() == "KBcNYbIC");
+
+    public override bool? IsApplicable(Game game)
+      // ReSharper disable once CompareOfFloatsByEqualityOperator
+    {
+      if (_perk == null) return false;
+      if (_perk.PrimaryBonus != 0.3f) return false;
```
Tyler-IN, line 58: I missed this. Why is this checked?

Issue #158, opened 2020-04-15
Nomadic Traditions doesn't work


Skau:
Riding perk.
Effect: `Increase the party speed bonus of your mounted troops by 30%`

Solganis:
This would be a game changer.

fnzr:
Solved with merge of #182

PR #182, opened 2020-04-18
Fix Nomadic Traditions Perk

fnzr:
Solves #158.
This gives a bigger buff than I was expecting, so if anyone wants to discuss the logic to make sure I did it right, I'd be happy to.


fnzr:
Oh, one important thing: the party speed is cached, and I cant force a refresh as it is, so after acquiring the Perk, you have to force a refresh by buying/selling a horse or upgrading/removing/adding troops of your army.
To fix this I'd have to add a listener to when a hero obtains a perk (`PerkActivationHandlerCampaignBehavior` maybe?) and try to invalidate the cache there. No patch yet added a campaign behavior, and might be on the edge of the scope of this project, so I didn't try it.
Should I?

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
```diff
@@ -0,0 +1,94 @@
...
+            troopsOnFoot += mobileParty2.Party.NumberOfMenWithoutHorse;
+            availableHorses += mobileParty2.ItemRoster.NumberOfMounts;
+          }
+        }
+
+        int mountedFootman = Math.Min(availableHorses, troopsOnFoot);
+        var mountedFootmenRatioMethod = typeof(DefaultPartySpeedCalculatingModel).GetMethod("GetMountedFootmenRatioModifier", BindingFlags.Instance | BindingFlags.NonPublic);
+        var defaultMountedFootmanRatio = (float)mountedFootmenRatioMethod.Invoke(__instance, new object[] {totalTroops, mountedFootman});
+        //default Mounted Footman constant is 0.3, so just apply it again
+        var perkMountedFootmanRatio = defaultMountedFootmanRatio;
```
tynakuh, line 78: Arguable if it should apply to these, i'd also expect the maths here to give me a new ratio of 0.09 not another flat 0.3

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
```diff
@@ -0,0 +1,94 @@
...
+        
+        var cavalryRatioMethod = typeof(DefaultPartySpeedCalculatingModel).GetMethod("GetCavalryRatioModifier", BindingFlags.Instance | BindingFlags.NonPublic);
+        var defaultCavalryRatio = (float)cavalryRatioMethod.Invoke(__instance, new object[] {totalTroops, mountedTroops});
+        //default Cavalry constant is 0.6, so do math
+        var perkCavalryRatio = Convert.ToSingle(0.3 * defaultCavalryRatio / 0.6);
```
tynakuh, line 83: The maths feels wrong here,  I don't think we should be dividing by 0.6 here. I'd expect that cavalry constant to go to 0.78f not 0.90f

tynakuh:
I'm all for adding a behaviour if it improves QoL/makes the game feel less buggy/weird, they should be pretty safe in general.

Skau:
There is a way you could do the math with the perk fields instead of the magic numbers. I see that the perk would have to be re-initialized because TW is wrongly initializing it with a primary bonus of `0.5f` when it should be `0.3f`. Its increment type (`AddFactor`) should be fine. 
How you have done it looks good to me and I'll approve it the way it's done when you double check the math that Tyna mentioned, but for consistency the perk field route is recommended.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
fnzr: I'm multiplying by 0.6 to get the flat cavalry ratio (see `GetCavalryRatioModifier` on the same class. Although at this point I should just have done the division manually)
yeah so both for both of these, I interpreted the perk as "gives 30% bonus on the flat ratio". So all said and done, with the perk active, total cavalry bonus would be `(0.6 + 0.3) * cavalryRatio` and mounted footman `(0.3 + 0.3) * mountedFootmanRatio`.
What you're suggesting is that I apply the 30% bonus AFTER the initial cut, so perk ratio is `0.3 * (0.6 + cavalryRatio)`? How did you get to 0.78f?

fnzr:
@Skau I'll see if I can make a patch to fix the Perk to the correct value then use it in this patch, should be straightforward.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
tynakuh: I'm thinking the 'constant' multiplier should be multiplied by 1.3, i.e. 0.6*1.3 = 0.78f  for the 30% bonus. but this is kind of semantics. so i'm happy for it to go either way.

fnzr:
Implemented changes suggested by @tynakuh because they make a lot of sense. So the result is: nomadic bonus = 0.3 * (footman on horse bonus +  cavalry bonus).

@Skau I pushed a fix to the NomadicTraditions perk object so if someone queries it the correct value is returned, but I didn't use it because I'd have to code all the small algebra in the comments, and that would be annoying.

Also added a campaign behavior to invalidate the party speed cache.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
```diff
@@ -0,0 +1,91 @@
...
+          }
+        }
+        float mountedFootman = Math.Min(availableHorses, troopsOnFoot);
+        var mountedFootmanRatio = totalTroops == 0 ? 0 : mountedFootman / totalTroops;
+        //MountedFootman magic number: 0.3f. Target total bonus: (0.3 * 1.3) = 0.39. Perk base ratio: 0.09
+        var perkMountedFootmanRatio = 0.09f * mountedFootmanRatio;
+        
+        var cavalryRatio = totalTroops == 0 ? 0 : mountedTroops / totalTroops;
+        //Cavalry magic number: 0.6f. Target total bonus: (0.6 * 1.3) = 0.78. Perk base ratio: 0.18
+        var perkCavalryRatio = 0.18f * cavalryRatio;
```
Tyler-IN, line 80: Like @Skau pointed out, can you extract the magic numbers (0.09f and 0.18f) to fields that calculate them from their constituents at patch load time?

If not at load time, resolve inline.

Tyler-IN: Look for examples in other patches using `ActivePerk` to refer to a cached PerkObject and extracting the PrimaryBonus and using Explainer's AddFactor to do the `(1 + 0.3) * n` magic.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsEffectPatch.cs
```diff
@@ -0,0 +1,18 @@
...
+  [HarmonyPatch(typeof(DefaultPerks), "InitializePerks")]
+  public class NomadicTraditionsEffectPatch {
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(DefaultPerks __instance) {
+      var nomadicTraditions = typeof(DefaultPerks).GetField("RidingNomadicTraditions", BindingFlags.Instance | BindingFlags.NonPublic).GetValue(__instance);
+      var primaryBonus = typeof(PerkObject).GetProperty("PrimaryBonus", BindingFlags.Instance | BindingFlags.Public);
+      primaryBonus.SetValue(nomadicTraditions, 0.3f);
+    }
+  }
```
Tyler-IN, line 17: Could you replace the perk's data following the pattern in the Disciplinarian patch instead of modifying InitializePerks?

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
fnzr: In this case, I cant rely on Explainer's AddFactor function because it's only set when the function is called by the user hovering over the horse icon in the map UI. If the party speed update is being made internally (or from parties not belonging to the MainHero), `explanation` is null.

fnzr:
Updated. I used `DefaultPerks.Riding.NomadicTraditions` instead of having a cached `_perk` field in the patch. Does it make a difference?
I don't understand why some patches use `PerkObject.FindFirst` method instead of just `DefaultPerks.X.Y`, since from what I've tested the results are the same.

Tyler-IN:
Yeah, I explained it to someone somewhere else - it's better to cache and grab the perk after initialization. Ping me on Discord and I'll try to shoot you the whole log, but basically after a load game or various other situations, due to serialization or some other means, the perks in DefaultPerks will not match the perks in the collection PerkObject.FindFirst accesses, Campaign.Current.PerkList. There are 2 different perk collections, and then there are the stand-alone references to the default perks - the only right one is Campaign.Current.PerkList.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
Tyler-IN: Explaination is null when the method is being called to get the value w/o an explanation, you (should) still use ExplainedNumber - it accepts null.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
```diff
@@ -0,0 +1,94 @@
...
+          }
+        }
+        float mountedFootman = Math.Min(availableHorses, troopsOnFoot);
+        var mountedFootmanRatio = totalTroops == 0 ? 0 : mountedFootman / totalTroops;
+        var perkMountedFootmanRatio = GetPerkBaseRatio(0.3f) * mountedFootmanRatio;
+        
+        var cavalryRatio = totalTroops == 0 ? 0 : mountedTroops / totalTroops;
+        var perkCavalryRatio = GetPerkBaseRatio(0.6f) * cavalryRatio;
+        
+        var baseSpeedMethod = typeof(DefaultPartySpeedCalculatingModel).GetMethod("CalculateBaseSpeedForParty", BindingFlags.Instance | BindingFlags.NonPublic);
```
Tyler-IN, line 85: This should be extracted to a static readonly field or get-only property.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
```diff
@@ -0,0 +1,94 @@
...
+          foreach (var mobileParty2 in mobileParty.AttachedParties) {
+            totalTroops += mobileParty2.MemberRoster.TotalManCount;
+            mountedTroops += mobileParty2.Party.NumberOfMenWithHorse;
+            troopsOnFoot += mobileParty2.Party.NumberOfMenWithoutHorse;
+            availableHorses += mobileParty2.ItemRoster.NumberOfMounts;
+          }
+        }
+        float mountedFootman = Math.Min(availableHorses, troopsOnFoot);
+        var mountedFootmanRatio = totalTroops == 0 ? 0 : mountedFootman / totalTroops;
+        var perkMountedFootmanRatio = GetPerkBaseRatio(0.3f) * mountedFootmanRatio;
```
Tyler-IN, line 80: Is this `.3` coming from `_perk.PrimaryBonus`?

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
```diff
@@ -0,0 +1,94 @@
...
+        
+        var cavalryRatio = totalTroops == 0 ? 0 : mountedTroops / totalTroops;
+        var perkCavalryRatio = GetPerkBaseRatio(0.6f) * cavalryRatio;
```
Tyler-IN, line 83: Is this `.6` coming from `_perk.PrimaryBonus` x 2?

Tyler-IN: If it's just the concern that you need to grab the perk multiple times, it's fine to copy the code to load the PerkObject in Reset, we'll probably refactor it to some per-load cache later.

src/CommunityPatch/Patches/Perks/Endurance/Riding/NomadicTraditionsPatch.cs
```diff
@@ -77,17 +77,20 @@ public sealed class NomadicTraditionsPatch : PatchBase<NomadicTraditionsPatch> {
...
+        var perkCavalryRatio = GetPerkBaseRatio(cavalryMagicNumber) * cavalryRatio;
 
-        var perkBonus = (perkCavalryRatio + perkMountedFootmanRatio) * baseSpeed;
+        var baseSpeed = (float)BaseSpeedMethodInfo.Invoke(__instance, new object[] {totalTroops});
+        
+        var explainedNumber = new ExplainedNumber(baseSpeed, explanation);
+        explainedNumber.AddFactor(perkCavalryRatio, ActivePatch._perk.Name);
+        explainedNumber.AddFactor(perkMountedFootmanRatio, ActivePatch._perk.Name);
+        var perkBonus = explainedNumber.ResultNumber - baseSpeed;
         __result += perkBonus;
```
fnzr: I'm not very confident is this. It does what I want and it gives the correct result, but is it the right way to go about it?

Tyler-IN: Can you take a screenshot of the explainer tooltip that this applies to?

I think since BaseSpeed is a const (CIL nomenclature; "static literal valuetype field") on the class, you might be able to declare a parameter that Harmony can extract for you; as `float ___BaseSpeed` instead of having to grab it by reflection.

In TW's implementation of `CalculatePureSpeed` you can see a pattern for this;
```cs
      ExplainedNumber bonuses = new ExplainedNumber(this.CalculateBaseSpeedForParty(num), explanation, (TextObject) null);
      float cavalryRatioModifier = this.GetCavalryRatioModifier(num, totalCavalryCount1);
      bonuses.AddFactor(cavalryRatioModifier, DefaultPartySpeedCalculatingModel._textCavalry);
...
return bonuses.ResultNumber;
```

`ExplainedNumber` may add a "Base" component to the explainer tooltip that you may need to remove; other patches do that, since you're using it to modify based on the base speed.
If you want to apply the factors manually, you can use Add and do the factor multiplication step separately against the `BaseSpeed`, so the first parameter to your `ExplainedNumber` can be `__result`, then you just `return __result`.

fnzr: Screenshot_37

PR #186, opened 2020-04-19
add perk 'heavier siege engines' from engineering tree and helper to organize and reuse code for next perks 

miguelcjalmeida:
This pull request adds a new perk to the game called 'Heavier siege engines', which should add 20% of bombardment damage against other siege engines. In order to do this, it was necessary to patch the 'SiegeEvent.BombardHitEngine', where the damage is applied to siege engine constructions. 

Besides the damage, it was also necessary to patch 'SandBox.ViewModelCollection.SandBoxUIHelper.GetSiegeEngineInProgressTooltip' to correctly display the perk buff and a tooltip updated with new damage values. Since there is a bunch of logic behind it, and same logic will repeat for at least three other perks, a new Helper class was created.

The SiegeTooltipHelper contains useful methods to deal with tooltip update, and also for easy patching the mentioned 'GetSiegeEngineInProgressTooltip'. 

There is also, in the last commit, the refactoring of Ballistic perk patch, which now will make use of this tooltip helper.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -7,6 +6,7 @@ using TaleWorlds.CampaignSystem;
 using TaleWorlds.Core;
 using TaleWorlds.Localization;
 using HarmonyLib;
+using Helpers;
 using TaleWorlds.Core.ViewModelCollection;
 using static CommunityPatch.HarmonyHelpers;
 
@@ -15,14 +15,9 @@ namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
   public sealed class BallisticsPatch : PatchBase<BallisticsPatch> {
 
     public override bool Applied { get; protected set; }
-    
-    private static readonly MethodInfo TooltipTargetMethodInfo =
-      Type.GetType("SandBox.ViewModelCollection.SandBoxUIHelper, SandBox.ViewModelCollection, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null")?
-        .GetMethod("GetSiegeEngineTooltip", BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly);
-    
-    private static readonly MethodInfo BombardTargetMethodInfo =
-      typeof(SiegeEvent).GetMethod("BombardHitEngine", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly);
 
+    private static readonly MethodInfo TooltipTargetMethodInfo = SiegeTooltipHelper.TargetMethodInfo;
+    private static readonly MethodInfo BombardTargetMethodInfo = typeof(SiegeEvent).GetMethod("BombardHitEngine", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly);
     private static readonly MethodInfo TooltipPatchMethodInfo = typeof(BallisticsPatch).GetMethod(nameof(TooltipPostfix), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.DeclaredOnly);
     private static readonly MethodInfo BombardPatchMethodInfo = typeof(BallisticsPatch).GetMethod(nameof(BombardPrefix), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.DeclaredOnly);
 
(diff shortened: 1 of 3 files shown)


src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
```diff
@@ -89,7 +75,7 @@ public override void Reset()
         _perk.Skill,
         (int) _perk.RequiredSkillValue,
         _perk.AlternativePerk,
-        SkillEffect.PerkRole.PartyLeader, _perk.PrimaryBonus,
+        SkillEffect.PerkRole.PartyLeader, 30f,
```
Tyler-IN, line 78: Went back to look - 30% is supposed to be `.3` if it's a factor increment type.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/HeavierSiegeEnginesPatch.cs
```diff
@@ -0,0 +1,123 @@
...
+      var parties = siegeEventSide.SiegeParties.Where(x => x.MobileParty != null);
+
+      foreach (var party in parties)
+      {
+        PerkHelper.AddPerkBonusForParty(perk, party.MobileParty, ref partyMemberRate);
+        PerkHelper.AddPerkBonusForParty(perk, party.MobileParty, ref partyMemberDamage);
+      }
+
+      bonusRateOnly = partyMemberRate.ResultNumber - 100;
+      bonusDamageOnly = partyMemberDamage.ResultNumber - baseDamage;
```
Tyler-IN, line 120: Is this a percentage or literal `100`? If it's part of a factor, it should be 1.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/HeavierSiegeEnginesPatch.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+      );
+      // most of the properties of skills have private setters, yet Initialize is public
+      _perk.Initialize(
+        textObjStrings[0],
+        textObjStrings[1],
+        _perk.Skill,
+        (int) _perk.RequiredSkillValue,
+        _perk.AlternativePerk,
+        _perk.PrimaryRole, 20f,
```
Tyler-IN, line 75: If this is 20%, it should be `.2` if it's a factor increment type.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/SiegeTooltipHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+      rangedDamageProperty.ValueLabel = amplifiedRangedDamage.ToString();
+    }
+    
+    public static void AddPerkTooltip(List<TooltipProperty> tooltips, PerkObject perk, float bonusValue) {
+      if (Math.Abs(bonusValue) < 0.05) return;
+      var isRate = perk.IncrementType == SkillEffect.EffectIncrementType.AddFactor;
+      var suffix = isRate ? "%" : "";
+      var tooltip = new TooltipProperty(perk.Name.ToString(), value: $"{bonusValue:F1}{suffix}", 0);
+      tooltips.Add(tooltip);
+    }
```
Tyler-IN, line 55: Are these working with percentages as numbers greater than 1? If so, that's probably not what we want, not what TW is going to do.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BallisticsPatch.cs
```diff
@@ -102,55 +88,46 @@ public override void Reset()
...
+
+      foreach (var party in parties)
+      {
+        PerkHelper.AddPerkBonusForParty(perk, party.MobileParty, ref partyMemberRate);
+        PerkHelper.AddPerkBonusForParty(perk, party.MobileParty, ref partyMemberDamage);
+      }
 
-      return PartiesHaveBallisticPerk(defenders);
+      bonusRateOnly = partyMemberRate.ResultNumber - 100;
+      bonusDamageOnly = partyMemberDamage.ResultNumber - baseDamage;
```
Tyler-IN, line 130: Is this a percentage or literal 100? If it's part of a factor, it should be 1.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/SiegeTooltipHelper.cs
miguelcjalmeida: The PerkHelper.AddToStat from TW multiply the perk bonus by 0.01, meaning that it expects the value to be 30 in case os 30%, not 0.30.

private static void AddToStat(ref ExplainedNumber stat, SkillEffect.EffectIncrementType effectIncrementType, float number, TextObject text)
	```
	private static void AddToStat(ref ExplainedNumber stat, SkillEffect.EffectIncrementType effectIncrementType, float number, TextObject text)
		{
			if (effectIncrementType == SkillEffect.EffectIncrementType.Add)
			{
				stat.Add(number, text);
				return;
			}
			if (effectIncrementType == SkillEffect.EffectIncrementType.AddFactor)
			{
				stat.AddFactor(number * 0.01f, text);
			}
		}
```
Should I create our own PerkHelper or keep using their style?

Tyler-IN: Keep using theirs.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/HeavierSiegeEnginesPatch.cs
Tyler-IN: This should be OK if the expected value is a whole value.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/HeavierSiegeEnginesPatch.cs
Tyler-IN: This should be ok if it runs through PerkHelpers.AddStat

PR #190, opened 2020-04-19
Make project easier to build for new people

iPherian:
Updates the readme to tell people the intended place for the repo. (the Modules dir).

Also reduce repetition in \*.csproj files by moving reference etc paths to project properties like 'MnB2BannerlordBinDir' when possible. While it works without this, someday someone might want to build with Bannerlord or the repo somewhere else and this would help them.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,8 @@
+<Project>
+
+    <PropertyGroup>
+        <MnB2BannerlordBinDir>../../../../bin/Win64_Shipping_Client</MnB2BannerlordBinDir>
+        <MnB2BannerlordLauncher>$(MnB2BannerlordBinDir)/TaleWorlds.MountAndBlade.Launcher.exe</MnB2BannerlordLauncher>
+    </PropertyGroup>
+
+</Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,7 +6,8 @@
 
         <!-- and not these -->
         <MnB2BannerlordBinDir>$(MnB2BannerlordDir)/bin/Win64_Shipping_Client</MnB2BannerlordBinDir>
-        <MnB2BannerlordLauncher>$(MnB2BannerlordBinDir)/TaleWorlds.MountAndBlade.Launcher.exe</MnB2BannerlordLauncher>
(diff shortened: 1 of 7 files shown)


fnzr:
I think making the project run out of the box is a great goal.
I think it's worth adding to the readme that the project expects it to be located under `Mount & Blade II Bannerlord\Modules` folder.
In the csproj file, we can leave the BinDir as is and change the launcher to variable to
`<MnB2BannerlordLauncher>./TaleWorlds.MountAndBlade.Launcher.exe</MnB2BannerlordLauncher>`
This way, in the default launch settings, instead of a new variable `$(MnB2BannerlordDir)`, we can use `$(MnB2BannerlordBinDir)` and the project should build and launch correctly.

iPherian:
Ohhhh you're supposed to put it in the Modules folder. That would've been a lot easier.

Tyler-IN:
There's also a `LocateSteamApp` project in there that can be wired up, but Facepunch.Steamworks is a little buggy inside MSBuild so I included custom-built binaries and a facade for dynamic linkage.
There's a usage of it over [here](https://github.com/ahmed007boss/Bannerlord-ModsSettings/blob/support-multiple-developers/src/Directory.Build.props);

```cs
<UsingTask TaskName="LocateSteamAppTask" AssemblyFile="$(MSBuildThisFileDirectory)\..\tools\LocateSteamApp.dll"/>
  <Target Name="LocateBannerlord" BeforeTargets="Build;Compile;CoreCompile;BeforeResolveReferences;ResolveAssemblyReferences">   
    <LocateSteamAppTask AppId="261550">
      <Output TaskParameter="Path" PropertyName="BannerlordDir"/>
    </LocateSteamAppTask>
...
```

But for some reason @ahmed007boss encounters an exception when it runs.

iPherian:
And then nobody would ever have to configure it. That sounds pretty cool.

Shall I do the easy part? Pushed commit.

Tyler-IN:
Does it work?

fnzr:
Doesnt'work out of the box for me, it's necessary to update `Properties\launchSettings.json`, `workingDirectory` key to `$(MnB2BannerlordBinDir)` and *csproj file `<MnB2BannerlordLauncher>./TaleWorlds.MountAndBlade.Launcher.exe</MnB2BannerlordLauncher>` for the launcher to show up.

iPherian:
@fnzr does that work better?

PR #192, opened 2020-04-20
Good materials perk from engineering tree

miguelcjalmeida:
This pull request adds the perk 'Good Materials', which increases your siege engines max health by 20%.

There are two commits to add the perk 'Good Materials' from engineering tree. The first one only deals with the damage application without any visual tooltips. The second one is supposed to implement the visuals, displaying the increased Max Health Points, which were very tricky to do due to lifecycle limitations when patching. 

**Commit 1**
The first commit have three patches. Since there are two routes to deploy a siege engine, one for player and another for AI, I had to patch them both. However, the player route required two patches to avoid adding references from 'SandBox.ViewModelCollection' into the project.

These two patches check whether the 'deployed siege engine list' has increased. If so, while the reserved siege engines list hasn't decreased, it means a new siege engine has just been deployed, allowing me to grab it by getting the last item of the list. Then it is possible to change its HP, but not the Max HP yet. Here comes the second commit.

**Commit 2** 
After the first commit, it was still necessary to update the Max Health Points from the just deployed siege engine. However, this property was not settable and neither its value stored in its own object. It was pointing to a list of predefined objects (siege engine types) that share their reference across many other siege engines, meaning changing the max hit points of one siege engine, would affect all the other siege engines of same type in the whole map. 

My approach was patching the Max Hit Points property of the siege engine (SiegeEngineConstructionProgress) to make use of a ConditionalWeakTable that would save their Max Hit Points. I chose this approach because it looked functional but I am really open to suggestions.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/SiegeTooltipHelper.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/SiegeTooltipHelper.cs
@@ -46,6 +46,16 @@ namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
       rangedDamageProperty.ValueLabel = amplifiedRangedDamage.ToString();
     }
     
+    public static void UpdateMaxHpTooltip(List<TooltipProperty> tooltips, float bonusHp)
+    {
+      var property = FindMaxHpTooltipProperty(tooltips);
+      if (property == null) return;
+
+      Double.TryParse(property.ValueLabel, out var currentHp);
+      var buffedHp = (int) (currentHp + bonusHp);
+      property.ValueLabel = buffedHp.ToString();
+    }
+    
     public static void AddPerkTooltip(List<TooltipProperty> tooltips, PerkObject perk, float bonusValue) {
       if (Math.Abs(bonusValue) < 0.05) return;
       var isRate = perk.IncrementType == SkillEffect.EffectIncrementType.AddFactor;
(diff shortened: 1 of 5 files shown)


Tyler-IN:
Don't forget to update README.md lol

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/SiegeTooltipHelper.cs
```diff
@@ -46,6 +46,16 @@ public static void UpdateRangedDamageTooltip(List<TooltipProperty> tooltips, flo
...
+      var property = FindMaxHpTooltipProperty(tooltips);
+      if (property == null) return;
+
+      Double.TryParse(property.ValueLabel, out var currentHp);
+      var buffedHp = (int) (currentHp + bonusHp);
+      property.ValueLabel = buffedHp.ToString();
+    }
+    
     public static void AddPerkTooltip(List<TooltipProperty> tooltips, PerkObject perk, float bonusValue) {
       if (Math.Abs(bonusValue) < 0.05) return;
```
Tyler-IN, line 60: can you comment on this? why 0.05?

miguelcjalmeida: Oh sorry, it doesn't have to be 0.05. Since ReSharper was complaining about the float check of bonusValue == 0f, it suggested me to use Math.Abs. The '0.05' could be any number close to zero. I am going to change it

Skau: You can check against epsilon

miguelcjalmeida: I did the checking with epsilon, but for some reason it was not returning the expected boolean everytime

Skau: Oh right. I see what you did instead, looks fine 👍

PR #193, opened 2020-04-20
add perk Armorcraft from Engineering tree

miguelcjalmeida:
I added the Armorcraft perk that increases character's armor by 10%. I had just to postfix the Agent.GetBaseArmorEffectivenessForBodyPart, to increase its output by 10% when the Agent had the perk active.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ArmorcraftPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ArmorcraftPatch.cs
@@ -1,0 +1,89 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using TaleWorlds.MountAndBlade;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
+  public class ArmorcraftPatch : PatchBase<ArmorcraftPatch> {
+    public override bool Applied { get; protected set; }
+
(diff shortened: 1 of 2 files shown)


Tyler-IN:

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ArmorcraftPatch.cs
```diff
@@ -0,0 +1,89 @@
...
+    // ReSharper disable once InconsistentNaming
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(ref float __result, ref Agent __instance) {
+      if (!__instance.IsHuman) return;
+      if (!__instance.IsHero) return;
+      if (!(__instance.Character is CharacterObject)) return;
+
+      var character = (CharacterObject)__instance.Character;
+      var hero = character?.HeroObject;
+      if (hero == null) return;
```
fnzr, line 82: There's no need for all these checks
```
if (__instance.IsHero && __instance.Character is CharacterObject charObj) {
 var hero = charObj.HeroObject;
 //apply perk
}
```
is enough.

Tyler-IN: I think at least the `IsHuman` check should probably be removed here. I don't know if any assertion that the character is human is valid for the armorcraft perk?

miguelcjalmeida: Without checking IsHuman first, the code crashes on the next line when checking if it is a hero. You could test this by simply jumping with your horse.

fnzr: Really? That's really unexpected. I wonder if there's an underlying issue on the TW code, a boolean test like that shouldn't crash.

PR #195, opened 2020-04-20
add Imperial Fire Perk from Engineering tree

miguelcjalmeida:
This pull request will add the perk 'Imperial Fire' which should enable fire siege engines. These fire engines were already enabled by default to every player. So what I had to do is removing those options from players that didn't have the perk active. 

There were two methods to be patched, one for the attacker team, the other for the defender team.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
@@ -1,0 +1,119 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using HarmonyLib;
+using Helpers;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
+
+  public class ImperialFirePatch : PatchBase<ImperialFirePatch> {
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo AttackerTargetMethodInfo =
+      Type.GetType("SandBox.ViewModelCollection.MapSiege.MapSiegeProductionVM, SandBox.ViewModelCollection, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null")?
+        .GetMethod("GetAllAttackerRangedMachines", BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly);
+    
+    private static readonly MethodInfo DefenderTargetMethodInfo =
+      Type.GetType("SandBox.ViewModelCollection.MapSiege.MapSiegeProductionVM, SandBox.ViewModelCollection, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null")?
+        .GetMethod("GetAllDefenderMachines", BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly);
+    
+    private static readonly MethodInfo PatchMethodInfoPostfix = typeof(ImperialFirePatch).GetMethod(nameof(Postfix), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.DeclaredOnly);
+
+    public override IEnumerable<MethodBase> GetMethodsChecked() {
+      yield return AttackerTargetMethodInfo;
+      yield return DefenderTargetMethodInfo;
+    }
+
+    private PerkObject _perk;
+
+    private static readonly byte[][] AttackerHashes = {
+      new byte[] {
+        // e1.1.0.225190
+        0xAC, 0x2D, 0x81, 0xE5, 0xC9, 0xA7, 0x97, 0x09,
+        0xE0, 0xD6, 0x9A, 0x22, 0xC8, 0x86, 0x28, 0x1F,
+        0x4A, 0x7B, 0x18, 0x9F, 0xCF, 0x27, 0xD8, 0xBC,
+        0x84, 0xD2, 0xD4, 0x6D, 0xC1, 0x30, 0x36, 0x43
+      }
+    };
+    
+    private static readonly byte[][] DefenderHashes = {
+      new byte[] {
+        // e1.1.0.225190
+        0x36, 0x8E, 0x5A, 0x15, 0x31, 0xDC, 0xB0, 0x6F,
+        0x0C, 0xD1, 0x55, 0xAF, 0x85, 0xC6, 0xD5, 0x8F,
+        0x60, 0x3F, 0x74, 0xDD, 0xF2, 0x4E, 0xC8, 0xD2,
+        0xC9, 0xAA, 0xEA, 0xE8, 0x66, 0x60, 0xE7, 0x57
+      }
+    };
+
+    public override void Reset()
+      => _perk = PerkObject.FindFirst(x => x.Name.GetID() == "UaZSa5mY");
+
+    public override bool? IsApplicable(Game game)
+      // ReSharper disable once CompareOfFloatsByEqualityOperator
+    {
+      if (_perk == null) return false;
+      if (_perk.PrimaryBonus != 0.3f) return false;
+      
+      var attackerPatchInfo = Harmony.GetPatchInfo(AttackerTargetMethodInfo);
+      if (AlreadyPatchedByOthers(attackerPatchInfo)) return false;
+      
+      var defenderPatchInfo = Harmony.GetPatchInfo(DefenderTargetMethodInfo);
+      if (AlreadyPatchedByOthers(defenderPatchInfo)) return false;
+
+      var attackerHash = AttackerTargetMethodInfo.MakeCilSignatureSha256();
+      var defenderHash = DefenderTargetMethodInfo.MakeCilSignatureSha256();
+      return attackerHash.MatchesAnySha256(AttackerHashes) && defenderHash.MatchesAnySha256(DefenderHashes);
+    }
+
+    public override void Apply(Game game) {
+      var textObjStrings = TextObject.ConvertToStringList(
+        new List<TextObject> {
+          _perk.Name,
+          _perk.Description
+        }
+      );
+      // most of the properties of skills have private setters, yet Initialize is public
+      _perk.Initialize(
+        textObjStrings[0],
+        textObjStrings[1],
+        _perk.Skill,
+        (int) _perk.RequiredSkillValue,
+        _perk.AlternativePerk,
+        _perk.PrimaryRole, 1f,
+        _perk.SecondaryRole, _perk.SecondaryBonus
+      );
+      if (Applied) return;
+
+      CommunityPatchSubModule.Harmony.Patch(AttackerTargetMethodInfo, postfix: new HarmonyMethod(PatchMethodInfoPostfix));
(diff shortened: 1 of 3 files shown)


Tyler-IN:

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,119 @@
...
+    // ReSharper disable once InconsistentNaming
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(ref IEnumerable<SiegeEngineType> __result) {
+      if (Hero.MainHero == null) return;
+      
+      var perk = ActivePatch._perk;
+      var partyMemberValue = new ExplainedNumber(0f);
+      PerkHelper.AddPerkBonusForParty(perk, Hero.MainHero.PartyBelongedTo, ref partyMemberValue);
+      
+      if (partyMemberValue.ResultNumber >= .99) return;
```
Tyler-IN, line 108: can you add comments to this? why is it .99?

miguelcjalmeida: I needed to check if float were equal or bigger than one, so I added a tolerance. This was confusing, so I refactored the code. Now there is a helper that do the comparison. Also I extracted that piece of code into a method to state its purpose more clearly.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,120 @@
...
+using TaleWorlds.Localization;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
+
+  public class ImperialFirePatch : PatchBase<ImperialFirePatch> {
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo AttackerTargetMethodInfo =
+      Type.GetType("SandBox.ViewModelCollection.MapSiege.MapSiegeProductionVM, SandBox.ViewModelCollection, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null")?
```
fnzr, line 19: This makes the mod dependent on Sandbox. Add `SandBox.ViewModelCollection.dll` to the assembly dependencies and use it to get the type instead of strings.
@Tyler-IN is there any problem with us depending on SandBox? For the load order stuff.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,120 @@
...
+        .GetMethod("GetAllAttackerRangedMachines", BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly);
+    
+    private static readonly MethodInfo DefenderTargetMethodInfo =
+      Type.GetType("SandBox.ViewModelCollection.MapSiege.MapSiegeProductionVM, SandBox.ViewModelCollection, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null")?
```
fnzr, line 23: Same as above.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    public override void Reset()
+      => _perk = PerkObject.FindFirst(x => x.Name.GetID() == "UaZSa5mY");
+
+    public override bool? IsApplicable(Game game)
+      // ReSharper disable once CompareOfFloatsByEqualityOperator
+    {
+      if (_perk == null) return false;
+      if (_perk.PrimaryBonus != 0.3f) return false;
```
fnzr, line 62: This stops your perk implementation if they fix the value of the it. I think it's best to break this in two patches, one for the Perk initialization fix, and other for the implementation of it, but it's up to you.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    // ReSharper disable once InconsistentNaming
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(ref IEnumerable<SiegeEngineType> __result) {
+      if (Hero.MainHero == null) return;
+      if (AnyPartyMemberHasThePerkActive(Hero.MainHero.PartyBelongedTo)) return;
+      __result = RemoveFireEngines(__result);
+    }
+    
+    private static bool AnyPartyMemberHasThePerkActive(MobileParty party) {
```
fnzr, line 107: Checking if anyone in the MobileParty has the perk is not enough. If the party belongs to an army, you have to check all parties in the army. See `party.Army`

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,120 @@
...
+      var perk = ActivePatch._perk;
+      var partyMemberValue = new ExplainedNumber(0f);
+      PerkHelper.AddPerkBonusForParty(perk, Hero.MainHero.PartyBelongedTo, ref partyMemberValue);
+      return partyMemberValue.ResultNumber.IsEqualOrBiggerThan(1f);
```
fnzr, line 111: Have you tried `partyMemberValue.ResultNumber.CompareTo(1f) >= 0`?

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
Tyler-IN: Using Type.GetType will work for always-loaded deps like this.

If the type init fails because Sandbox isn't loaded, the patch will be skipped.

This keeps loose coupling. It's okay.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
Tyler-IN: There are several perks like this. It might make sense for the sake of time to do that as part of a refactor, given the volume of patches he's making.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
miguelcjalmeida: Could you review again please?

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -107,7 +107,12 @@ public override void Reset()
     private static bool AnyPartyMemberHasThePerkActive(MobileParty party) {
       var perk = ActivePatch._perk;
       var partyMemberValue = new ExplainedNumber(0f);
-      PerkHelper.AddPerkBonusForParty(perk, Hero.MainHero.PartyBelongedTo, ref partyMemberValue);
+      PerkHelper.AddPerkBonusForParty(perk, party, ref partyMemberValue);
+
+      if (party.Army?.Parties != null)
+        foreach (var armyParty in party.Army?.Parties)
+          PerkHelper.AddPerkBonusForParty(perk, armyParty, ref partyMemberValue);
```
fnzr, line 114: The function `AddPerkBonusForParty` is being called twice for the initial party, once with `party`, and once inside the foreach loop.

miguelcjalmeida: Since the perk is a enable/disable feature, I think that is ok, or no?

fnzr: I don't think so, since the enable/disable condition is `partyMemberValue.ResultNumber.IsEqualOrBiggerThan(1f);`, if you add the same factor twice it will affect the result.

fnzr: Ah, I see it now, you changed the Perk primaryValue to 1. I guess it doesn't make a difference then, but if you don't need the primaryValue and just a boolean, I think it would be much clearer if you wrote
```
if (party.Army == null) 
    return party.LeaderHero.GetPerkValue(__perk);
foreach (var armyParty in party.Army.Parties) {
    if (armyParty.LeaderHero.GetPerkValue(__perk)
        return true;
return false;
```
instead of taking using `PerkHelper` to add a bunch of 1s.

miguelcjalmeida: I believe the code you wrote don't check the party members of the clans. I just used the PerkHelper to make that job easier.

fnzr: I see. Well, it's fine.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,125 @@
...
+using Helpers;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
+
+  public class ImperialFirePatch : PatchBase<ImperialFirePatch> {
+    public override bool Applied { get; protected set; }
```
iPherian, line 16:
```suggestion
    // used as a kind of boolean, 1 is on, Perk system default is 0 which is off
    private const float PatchPerkBonus = 1f;

    public override bool Applied { get; protected set; }
```

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+      );
+      // most of the properties of skills have private setters, yet Initialize is public
+      _perk.Initialize(
+        textObjStrings[0],
+        textObjStrings[1],
+        _perk.Skill,
+        (int) _perk.RequiredSkillValue,
+        _perk.AlternativePerk,
+        _perk.PrimaryRole, 1f,
```
iPherian, line 89:
```suggestion
        _perk.PrimaryRole, PatchPerkBonus,
```

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ImperialFirePatch.cs
```diff
@@ -0,0 +1,125 @@
...
+    private static bool AnyPartyMemberHasThePerkActive(MobileParty party) {
+      var perk = ActivePatch._perk;
+      var partyMemberValue = new ExplainedNumber(0f);
+      PerkHelper.AddPerkBonusForParty(perk, party, ref partyMemberValue);
+
+      if (party.Army?.Parties != null)
+        foreach (var armyParty in party.Army?.Parties)
+          PerkHelper.AddPerkBonusForParty(perk, armyParty, ref partyMemberValue);
+
+      return partyMemberValue.ResultNumber.IsEqualOrBiggerThan(1f);
```
iPherian, line 116:
```suggestion
      return partyMemberValue.ResultNumber.IsEqualOrBiggerThan(PatchPerkBonus);
```

PR #196, opened 2020-04-20
add perk wall breaker and tooltip for ranged wall damage

miguelcjalmeida:
Add the 'wall breaker' perk that increases damage to walls by 25%. Since from now on the damage to walls differ from the damage to siege engines, I have added a new tooltip property so the player can clearly see how much damage he can apply to each target. 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ This is going to be a mod that just fixes up some things in Mount &amp; Blade 2:
     * Everyday Engineer
     * Builder
     * Armorcraft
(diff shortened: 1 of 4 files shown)


Tyler-IN:
Nice screenshot

Tyler-IN:
Needs to be rebased due to README.md, this will be a repeating theme.
It's not your fault other PRs got merged. At some point I'm just going to establish a policy that the person in charge of merging just does it after all needed approvals.
Good work. 👍

PR #197, opened 2020-04-20
add the Resolute perk from engineering tree

miguelcjalmeida:
Adds the 'resolute' perk which is supposed to reduce morale loss during sieges. 

This patch grabs every negative effect on morale and reduce their impact by a rate. The rate is calculated based on the party members that have the perk activated. The bonus is 50%, however, they are not additive, meaning they stack multiplicatively. 

So if the player is having a -30 morale debuff because of starvation, one member having this perk would reduce the effect to -15 (50%). Two members would reduce by 75% instead.

The morale loss reduction effects apply only when the perk is active and the character party is on siege.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
@@ -14,7 +14,8 @@ namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
   public class ResolutePatch : PatchBase<ResolutePatch> {
     public override bool Applied { get; protected set; }
 
-    private static readonly MethodInfo TargetMethodInfo = typeof(DefaultPartyMoraleModel).GetMethod(nameof(DefaultPartyMoraleModel.GetEffectivePartyMorale), BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
+    private static readonly Type DefaultPartyMoraleModelType = typeof(DefaultPartyMoraleModel);
+    private static readonly MethodInfo TargetMethodInfo = DefaultPartyMoraleModelType.GetMethod(nameof(DefaultPartyMoraleModel.GetEffectivePartyMorale), BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
     private static readonly MethodInfo PatchMethodInfoPostfix = typeof(ResolutePatch).GetMethod(nameof(Postfix), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.DeclaredOnly);
     public override IEnumerable<MethodBase> GetMethodsChecked() {
       yield return TargetMethodInfo;
@@ -35,11 +36,9 @@ namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
       }
     };
     
-    // ReSharper disable once CompareOfFloatsByEqualityOperator
     public override bool? IsApplicable(Game game)
     {
       if (_perk == null) return false;
-      if (_perk.PrimaryBonus != 0.3f) return false;
       
       var patchInfo = Harmony.GetPatchInfo(TargetMethodInfo);
       if (AlreadyPatchedByOthers(patchInfo)) return false;
@@ -73,18 +72,18 @@ namespace CommunityPatch.Patches.Perks.Intelligence.Engineering {
     }
     
     [MethodImpl(MethodImplOptions.NoInlining)]
-    public static void Postfix(ref float __result, MobileParty mobileParty, StatExplainer explanation = null) {
+    public static void Postfix(ref float __result, ref DefaultPartyMoraleModel __instance, MobileParty mobileParty, StatExplainer explanation = null) {
       var perk = ActivePatch._perk;
       
       if (mobileParty.SiegeEvent == null) return;
       var partyMemberBonus = GetPartyMemberBonus(mobileParty, perk);
-      if (partyMemberBonus < .01) return;
+      if (partyMemberBonus.IsEqualOrLesserThan(0f)) return;
 
       var partyMembersCount = (int) (partyMemberBonus / perk.PrimaryBonus);
       var multiplicativeBonus = CalculateMultiplicativeBonus(perk.PrimaryBonus / 100, partyMembersCount);
       var moraleLossReductionRate = multiplicativeBonus * -1;
       
-      var totalMoraleLossReduction = CalculateTotalMoraleLossReduction(mobileParty, moraleLossReductionRate);
+      var totalMoraleLossReduction = CalculateTotalMoraleLossReduction(__instance, mobileParty, moraleLossReductionRate);
       explanation?.AddLine(perk.Name.ToString(), totalMoraleLossReduction);
       __result += totalMoraleLossReduction;
     }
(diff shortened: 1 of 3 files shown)


src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
```diff
@@ -0,0 +1,142 @@
...
+      if (oversize > 0) return -1f * (float) Math.Sqrt(oversize);
+      return 0f;
+    }
+
+    private static float CalculateFoodVarietyPenalty(MobileParty party) {
+      var variety = party.ItemRoster.FoodVariety;
+      if (variety <= 1) return -2;
+      if (variety <= 2) return -1;
+      return 0;
+    }
```
fnzr, line 140: Use the existing `CalculateFoodVarietyMoraleBonus` and `GetPartySizeMoraleEffect` functions instead.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      return finalRate;
+    }
+
+    private static int CalculateTotalMoraleLossReduction(MobileParty mobileParty, float moraleLossReductionRate)
+    {
+      var recentEventsMoraleLossReduction = Math.Min(mobileParty.RecentEventsMorale, 0) * moraleLossReductionRate;
+      var starvationMoraleLossReduction = mobileParty.Party.IsStarving ? -30 * moraleLossReductionRate : 0;
+      var unpaidWageMoraleLossReduction = mobileParty.HasUnpaidWages > 0f ? -20 * moraleLossReductionRate : 0;
```
fnzr, line 117: Use the existing `GetStarvationMoralePenalty` and `GetNoWageMoralePenalty` functions instead of hardcoded numbers

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
miguelcjalmeida: Just to clarify my understanding, is your suggestion for me to patch the following property/methods ? 
- MobileParty.RecentEventsMorale
- DefaultPartyMoraleModel.GetStarvationMoralePenalty()
- DefaultPartyMoraleModel.GetNoWageMoralePenalty()
- DefaultPartyMoraleModel.CalculateFoodVarietyMoraleBonus()
- DefaultPartyMoraleModel.GetPartySizeMoraleEffect()

What do you suggest me to do regarding the display of the Party Morale tooltip? I think it is good when we let the player know when an effect is active. Currently, I was displaying all the morale loss reduction sum up like this:

Starvation: -30
No wage: -20
Recent Events: -6
Resolute: +28.
----

By implementing the resolute perk on each property/method individually, I guess the final result look like this:
Starvation: -30
Resolute: +15
No Wage: -20
Resolute: +10
Recent events: -6
Resolute: +3

Is this acceptable or should I think hard about an alternative?

fnzr: StatExplainer handles calls of `AddLine` with the same `name` parameter. So if you call `explanation.AddLine("Resolute", 15f, StatExplainer.OperationType.Add)` and later `explanation.AddLine("Resolute", 3f, StatExplainer.OperationType.Add)`, when the text actually shows, it will be `Resolute: +18`.

miguelcjalmeida: Awesome :)

miguelcjalmeida: How would you access statExplainer from within MobileParty.RecentEventsMorale?

I guess patching DefaultPartyMoraleModel.GetEffectivePartyMorale is still necessary, right?

fnzr: hmm yeah, tricky. You could create a `_statExplainer` field like we do in other patches with `_perk`, but it's probably better to patch `GetEffectivePartyMorale`. Depends on how the code is looking, but either way is fine.

miguelcjalmeida: I was working on refactoring the class but I noticed the methods DefaultPartyMoraleModel.GetStarvationMoralePenalty and DefaultPartyMoraleModel.GetNoWageMoralePenalty also don't receive the statExplainer. Do you still think it is worthy implementing a patch the other methods ?

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
miguelcjalmeida: I'll have it done

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
fnzr: That was an option. Another one is patch only `GetEffectivePartyMorale`, but make sure to get the values of PartySizeMorale, FoodVariety, etc by calling the respective functions instead of rewriting them on the patch.

src/CommunityPatch/Patches/Perks/Intelligence/Engineering/ResolutePatch.cs
```diff
@@ -0,0 +1,165 @@
...
+    private static float GetMoralePenaltyFromMethod(DefaultPartyMoraleModel model, string methodName, MobileParty party) {
+      var method = DefaultPartyMoraleModelType.GetMethod(methodName, BindingFlags.Instance | BindingFlags.DeclaredOnly | BindingFlags.NonPublic);
+      if (method == null) return 0f;
+      
+      var moraleBonus = (int) method.Invoke(model, new object[] {party});
+      return GetMoralePenalty(moraleBonus);
+    }
+    
+    private static float GetMoralePenaltyFromMethodWithExplainedNumber(DefaultPartyMoraleModel model, string methodName, MobileParty party) {
+      var method = DefaultPartyMoraleModelType.GetMethod(methodName, BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.DeclaredOnly);
```
fnzr, line 150: Instead of passing the method name as string every call, this function could receive a `MethodInfo` and just invoke it. So you'd declere the MethodInfo just once:
`private static readonly GetNoWageMoralePenaltyMethodInfo = DefaultPartyMoraleModelType.GetMethod(methodName, BindingFlags.Instance | BindingFlags.DeclaredOnly | BindingFlags.NonPublic);`
and then the method signature would be `GetMoralePenaltyFromMethodWithExplainedNumber(DefaultPartyMoraleModel model, MethodInfo method, MobileParty party)` and you could just `method.Invoke(model, new object[] {party})`.
All this is doing is preventing a `GetMethod` call every time this function is called, so I'm not sure if it's worth it.
What do you think, @Tyler-IN?

Tyler-IN: Yeah, need to extract as much reflection out to members, either static or instance.

Tyler-IN: Will do this on dev. Need to get an RC going.

PR #198, opened 2020-04-20
add perk Builder from engineering tree

miguelcjalmeida:
This perk increases building speed. I check whether the town has a governor with the perk active, then I increase its production power by 50%.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BuilderPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BuilderPatch.cs
@@ -1,0 +1,109 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using HarmonyLib;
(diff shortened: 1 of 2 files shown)


src/CommunityPatch/Patches/Perks/Intelligence/Engineering/BuilderPatch.cs
```diff
@@ -0,0 +1,109 @@
...
+
+      var perk = ActivePatch._perk;
+      var productionPowerBonus = new ExplainedNumber(productionPower, explanation);
+      productionPowerBonus.AddFactor(perk.PrimaryBonus, perk.Name);
+      
+      productionPower = (int) productionPowerBonus.ResultNumber;
+    }
+    
+    private static bool HasGovernorWithBuilderPerk(Town town)
+      => town.Governor?.GetPerkValue(ActivePatch._perk) == true;
```
fnzr, line 107: nitpick, but is this function necessary?

Tyler-IN: It may get refactored out later. It does add descriptiveness to the intent. Renaming `_perk` might have the same effect on readability.

PR #199, opened 2020-04-20
add perk everyday engineer from engineering three

miguelcjalmeida:
Add the perk 'everyday engineer' which should improve all default town projects by 60%. 

There is a method that is called to get the effect power of every project. I patched this method, checking if its type is 'DefaultProject' before applying the perk bonus. 

Since all default town projects have an effect power of 1, and the value type is integer, the perk is actually doubling the effect power instead of increasing by 60%.

### DIFF
--- a/src/CommunityPatch/FloatHelper.cs
+++ b/src/CommunityPatch/FloatHelper.cs
@@ -3,8 +3,16 @@ using System.Runtime.CompilerServices;
 
 namespace CommunityPatch {
   public static class FloatHelper {
+
+    private const float Tolerance = 0.0001f;
+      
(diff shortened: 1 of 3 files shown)


src/CommunityPatch/FloatHelper.cs
```diff
@@ -3,8 +3,16 @@
 
 namespace CommunityPatch {
   public static class FloatHelper {
+
+    private const float Tolerance = 0.0001f;
```
fnzr, line 7: Is this value arbitrary?

fnzr: Nevermind, read the discord conversation. All good.

PR #211, opened 2020-04-22
add the perk Scavenger from roguery

miguelcjalmeida:
This PR adds the perk scavenger, which increases loot by 25%. There is a internal class LootCollector responsible for distributing loot among parties. It already checked for the roguery skill level to increase the given loot. So what I did is a prefix patch that would make the same calls, but without modifing the object state, to reward only the extra loot according to the perk bonus.


src/CommunityPatch/Patches/Perks/Cunning/Roguery/ScavengerPatch.cs
```diff
@@ -0,0 +1,132 @@
...
+    
+    private static List<TroopRosterElement> GetCasualtiesShare(float lootAmount, TroopRoster casualtiesInBattle) {
+      var casualtiesShare = new List<TroopRosterElement>();
+      
+      for (var i = casualtiesInBattle.Count - 1; i >= 0; i--) {
+        var troopSize = casualtiesInBattle.GetElementNumber(i);
+        var troop = casualtiesInBattle.GetCharacterAtIndex(i);
+        var emptyTroop = new TroopRosterElement(troop);
+        
+        for (var num4 = 0; num4 < troopSize; num4++) {
```
Tyler-IN, line 100:
```suggestion
        for (var trooper = 0; trooper < troopSize; trooper++) {
```

PR #212, opened 2020-04-22
Add bannerlord dependency on StoryMode

iPherian:
Use of it in [EarlyStoryVisibleTimeoutPatch](https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/blob/dev/src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs#L75). ( StoryMode.StoryModePhases.FirstPhase, etc )

### DIFF
--- a/SubModule.xml
+++ b/SubModule.xml
@@ -7,7 +7,9 @@
     <SingleplayerModule value="true" />
     <MultiplayerModule value="true" />
     <Official value="false" />
-    <DependedModules />
+    <DependedModules>
+        <DependedModule Id="StoryMode" />
+    </DependedModules>
     <SubModules>
         <SubModule>
             <Name value="CommunityPatch" />


Tyler-IN:
I don't want the type scanner in the loader (the ones that looks for DotNetObject and ManagedObject) or the assembly preloader to get the wrong idea.

As we've done for SandBox, we use Type.GetType to implicitly grab the type reference.
This means it will fail if for some reason SandBox is not loaded or the symbol changes instead of failing any sort of validation imposed by user or runtime.

Tyler-IN:
See alternative; https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/pull/218

iPherian:
This one's definitely not needed then.

PR #214, opened 2020-04-22
add perk slip into shadows from roguery

miguelcjalmeida:
Add the scout perk 'Slip into shadows' which reduces the chances of being caught when sneaking into towns by 20%. It seems the sneaking is a player only functionality.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs
+++ b/src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs
@@ -1,0 +1,64 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.SandBox.GameComponents.Map;
+using TaleWorlds.Core;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Cunning.Roguery {
+
+  public sealed class SlipIntoShadows : PatchBase<SlipIntoShadows> {
+    
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo TargetMethodInfo = typeof(DefaultDisguiseDetectionModel).GetMethod(nameof(DefaultDisguiseDetectionModel.CalculateDisguiseDetectionProbability), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly);
+    private static readonly MethodInfo PatchMethodInfo = typeof(SlipIntoShadows).GetMethod(nameof(Postfix), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.DeclaredOnly);
+
+    public override IEnumerable<MethodBase> GetMethodsChecked() {
+      yield return TargetMethodInfo;
+    }
+
+    private PerkObject _perk;
+
+    private static readonly byte[][] Hashes = {
+      new byte[] {
+        // e1.1.0.225190
+        0xAD, 0x2A, 0x34, 0x12, 0x4D, 0x03, 0xF5, 0x3D,
+        0x12, 0xE9, 0x1C, 0xFE, 0x07, 0x49, 0xFB, 0x75,
+        0xB2, 0x55, 0xC1, 0xA9, 0xF0, 0xB7, 0xF9, 0xBB,
+        0x1F, 0x00, 0xF2, 0x7F, 0x04, 0x77, 0x84, 0x8E
+      }
+    };
+
+    public override void Reset()
+      => _perk = PerkObject.FindFirst(x => x.Name.GetID() == "Eth2Z6qK");
+
+    public override bool? IsApplicable(Game game) {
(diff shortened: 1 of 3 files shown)


src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Runtime.CompilerServices;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.SandBox.GameComponents.Map;
+using TaleWorlds.Core;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Cunning.Roguery {
+
+  public sealed class SlipIntoShadows : PatchBase<SlipIntoShadows> {
```
fnzr, line 12: Add `Patch` as suffix for both class name and file, to follow the same pattern as the others patches in the mod

src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs
```diff
@@ -0,0 +1,64 @@
...
+    public override void Apply(Game game) {
+      if (Applied) return;
+      CommunityPatchSubModule.Harmony.Patch(TargetMethodInfo, postfix: new HarmonyMethod(PatchMethodInfo));
+      Applied = true;
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(ref float __result) {
+      var perk = ActivePatch._perk;
+      var scout = Hero.MainHero.PartyBelongedTo.EffectiveScout;
```
fnzr, line 58: If the MainHero has the perk but is not the scout of the party, the perk wouldn't be applied. Is this right?

fnzr: I guess it is, since the Perk is initialized with PerkRole.Scout.

src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs
```diff
@@ -0,0 +1,64 @@
...
+      if (scout?.GetPerkValue(perk) != true) return;
+      
+      __result -= perk.PrimaryBonus;
```
wonkotron, line 61: It looks like you might be adding to the effective chance of being detected instead of subtracting, since the perk bonus is already negative:

```this.RoguerySlipIntoShadows.Initialize("{=Eth2Z6qK}Slip Into Shadows", "{=Z1rfW4jL}-20% chance of being caught while sneaking into towns.", DefaultSkills.Roguery, this.GetTierCost(5), this.RogueryConcealedBlade, SkillEffect.PerkRole.Scout, -0.2f, SkillEffect.PerkRole.None, 0f, SkillEffect.EffectIncrementType.AddFactor);```

Comment (miguelcjalmeida) [reply src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs:61]: No, I am subtracting __result with -=

File: src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs (hunk 1 above)
Comment (miguelcjalmeida) [reply src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs:12]: Sorry! I'll have that fixed

File: src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs (hunk 3 above)
Comment (wonkotron) [reply src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs:61]: The only reason I ask is that the method you're modifying is CalculateDisguiseDetectionProbability, which does not imply chance of sneaking undetected but rather the opposite.

Comment (miguelcjalmeida) [reply src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs:61]: Just to make sure I double checked, and we are fine. The name is very confusing as you pointed out.

Comment (Tyler-IN) [reply src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlipIntoShadows.cs]: From `EncounterGameMenuBehavior` where it is used;
```cs
		// Token: 0x060022C4 RID: 8900 RVA: 0x00099A4C File Offset: 0x00097C4C
		private bool game_menu_town_disguise_yourself_on_condition(MenuCallbackArgs args)
		{
			args.optionLeaveType = GameMenuOption.LeaveType.Submenu;
			MBTextManager.SetTextVariable("SNEAK_CHANCE", Math.Round((double)(Campaign.Current.Models.DisguiseDetectionModel.CalculateDisguiseDetectionProbability(Settlement.CurrentSettlement) * 100f)), false);
			return this._accessDetails.AccessLevel == SettlementAccessModel.AccessLevel.LimitedAccess && this._accessDetails.LimitedAccessSolution == SettlementAccessModel.LimitedAccessSolution.Disguise;
		}
```

The positive output of CalculateDisguiseDetectionProbability does indicate stealth instead of detection. Yep, kind of dumb.

wonkotron: I agree.  Function name is misleading, but @miguelcjalmeida 's patch looks correct.

PR #215, opened 2020-04-22
Fix Aserai Cheap Caravans cultural feat

wonkotron:
-  Tested creating a caravan with Aserai character:
image
- Tested creating a caravan with non-Aserai character, result is no refund and no flavor text


src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
```diff
@@ -0,0 +1,87 @@
...
+                    // As of e1.2.0.226271, Aserai bonus is not set
+                    // Assume 30% caravan discount per character creation screen text
+                    creditAmount = (int) (0.30f * (__state - Hero.MainHero.Gold));
+                }
+                else {
+                    // If bonus is set, use it
+                    creditAmount = (int) ((1.0f - DefaultFeats.Cultural.AseraiCheapCaravans.EffectBonus) * (__state - Hero.MainHero.Gold));
+                }
+
+                var flavorText = "An Aserai clerk nods to you and slips a coin purse into your hands.";
```
Tyler-IN, line 78:
```suggestion
                var flavorText = new TextObject("{=AseraiCheapCaravansFlavorText}An Aserai clerk nods to you and slips a coin purse into your hands.").ToString();
```

Tyler-IN: We can add a localization xml file later if needed.

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
```diff
@@ -0,0 +1,87 @@
...
+            // of the target function call
+            if (Hero.MainHero.Gold < __state &&
+                Hero.MainHero.IsHumanPlayerCharacter &&
+                Hero.MainHero.CharacterObject.GetFeatValue(DefaultFeats.Cultural.AseraiCheapCaravans)) {
+                
+                int creditAmount = 0;
+                if (DefaultFeats.Cultural.AseraiCheapCaravans.EffectBonus == 0f) {
+                    // As of e1.2.0.226271, Aserai bonus is not set
+                    // Assume 30% caravan discount per character creation screen text
+                    creditAmount = (int) (0.30f * (__state - Hero.MainHero.Gold));
```
iPherian, line 71: Could be cool if you tried to parse it from the description, then fallback on 30%
```suggestion
                    // try to get it from the feat description, or assume 30% discount as per what the
                    // character creation screen text at one time said.
                    float discount = 0.30f;
      	            try {
                        Match featDescMatch = Regex.Match(DefaultFeats.Cultural.AseraiCheapCaravans.Description.ToString(),
                                                          @"Caravans are ([\d\.]+)% cheaper to build", RegexOptions.IgnoreCase);
      	                if (featDescMatch.Success) { 
							              discount = float.Parse(featDescMatch.Groups[1].Value, CultureInfo.InvariantCulture) / 100.0f;
                        }
                    } catch (FormatException) {}
                    creditAmount = (int) (discount * (__state - Hero.MainHero.Gold));
```

iPherian: would also require at the top of file:

```
using System.Globalization;
using System.Text.RegularExpressions;
```

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
```diff
@@ -0,0 +1,87 @@
...
+                }
+                else {
+                    // If bonus is set, use it
+                    creditAmount = (int) ((1.0f - DefaultFeats.Cultural.AseraiCheapCaravans.EffectBonus) * (__state - Hero.MainHero.Gold));
```
iPherian, line 75: If they did set it to 30% someday, wouldn't this result in you getting a 1.0 - 0.3 = 70% discount instead of 30%? Unless they changed it in the beta, the other numbers in the feats file seem to be the modifier itself after all.

_TaleWorlds.CampaignSystem.DefaultFeats, TaleWorlds.CampaignSystem.dll, game e1.1.2_:

```
      ..., "{=ByrkbDoK}10% extra speed bonus for horsemen on campaign map.", 0.1f, ...
      ..., "{=Impk9uLu}20% less speed penalty from snow.", 0.2f, ...
```

```suggestion
                    creditAmount = (int) (DefaultFeats.Cultural.AseraiCheapCaravans.EffectBonus * (__state - Hero.MainHero.Gold));
```

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
Tyler-IN: Is Description a TextObject? ToString will localize it if it's not English.

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
Tyler-IN: As a factor, it makes sense to be .7 ... The IsApplicable check could verify the EffectBonus is .7

wonkotron: @iPherian , this is a bug in my code.  Will fix

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
iPherian: > This .3 doesn't exist somewhere else, like the Perk's primary bonus?
> `if (1 - DefaultFeats.Cultural.AseraiCheapCaravans.EffectBonus)` gives the .3, that should be expressed as a property in the patch and reused, most likely.

This is inside an `if` branch where the [EffectBonus is 0](https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/pull/215/files/93fb699b2b7fb454abba5a9362306037594b0e95#diff-455508814d7cc07863450716b5a3e047R68) which he assumed to be wrong and substituted a different value. Indeed if the [EffectBonus exists](https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/pull/215/files/93fb699b2b7fb454abba5a9362306037594b0e95#diff-455508814d7cc07863450716b5a3e047R75) it is used.

> Is Description a TextObject? ToString will localize it if it's not English.
> 

Yes Description is a TextObject. I'll see if I can get the raw text to avoid the localization, but it's only a fallback. Code uses EffectBonus if available.

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
iPherian: Admittedly i'm not absolutely certain. Here is the entire feat code from the base game. Happy to let others interpret what it means:

```
private void InitializeAll()
    {
      this._cultureBattanianForestAgility.Initialize("{=g4jzOBYX}Battanian Forest Agility", "{=5axKTUpX}Forests give 10% less speed penalty to parties", 0.1f, FeatObject.AdditionType.AddFactor);
      this._cultureVlandianBattlePerception.Initialize("{=2pzjOTtj}Vlandian Battle Perception", "{=bL5YPUbd}20% more upgrade XP to troops from battles.", 0.2f, FeatObject.AdditionType.AddFactor);
      this._cultureKhuzaitCavalryAgility.Initialize("{=egNhD6fF}Khuzait Cavalry Agility", "{=ByrkbDoK}10% extra speed bonus for horsemen on campaign map.", 0.1f, FeatObject.AdditionType.AddFactor);
      this._cultureSturgianSnowAgility.Initialize("{=lMIi4CXO}Sturgian Snow Agility", "{=Impk9uLu}20% less speed penalty from snow.", 0.2f, FeatObject.AdditionType.AddFactor);
      this._cultureAseraiCheapCaravans.Initialize("{=NJmTcbOI}Aserai Cheap Caravans", "{=7kGGgkro}Caravans are 30% cheaper to build. 10% less trade penalty.", 0.0f, FeatObject.AdditionType.AddFactor);
      this._cultureEmpireConstructionExpert.Initialize("{=b7W48wVD}Empire Construction Expert", "{=GIljHGh9}20% construction speed bonus to town projects, wall repairs and siege engines.", 0.2f, FeatObject.AdditionType.AddFactor);
    }
```

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
iPherian: Sadly, looking over TextObject.cs, I don't see a way to avoid the localization.

wonkotron: The 0.3f bonus isn't defined anywhere else at the moment for this feat

iPherian: Although you could get the unlocalized string from the TextObject with reflection. But i'll leave this optional for now. It's just a nifty little thing.

Tyler-IN: There's a trick I've done elsewhere, but I don't think we should use it.
```cs
      var textObjStrings = TextObject.ConvertToStringList(
        new List<TextObject> {
          _perk.Name,
          _perk.Description
        }
      );
```
This extracts `Value` field values directly from TextObjects.

Tyler-IN:
Woops, waiting on another commit I guess.

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
iPherian: Cool, i'll remember that. 👍

(when there's no other choice)

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansPatch.cs
iPherian: [added](https://github.com/Tyler-IN/MnB2-Bannerlord-CommunityPatch/pull/215/commits/cca4533b317e3af104869ad8b2a4982095a281f8#diff-455508814d7cc07863450716b5a3e047R61)

src/CommunityPatch/Patches/Feats/AseraiCheapCaravansEffectBonusPatch.cs
```diff
@@ -0,0 +1,65 @@
...
+                Applied = false;
+            }
+            else {
+                featObject.Initialize(
+                    featObject.Name.ToString(),
+                    featObject.Description.ToString(),
+                    PatchedEffectBonusValue,
+                    featObject.IncrementType);
+
+                Applied = !featObject.EffectBonus.IsDifferentFrom(PatchedEffectBonusValue);
```
wonkotron, line 57: Should an exception be thrown here if Applied is false?  What is the standard for logging errant behavior?

iPherian: My admittedly limited experience here says yes throw an exception. That way, the mod will write an error message into the log, and it's a way to ensure that the mod's loading code is aware of the fact that this patch failed.

Tyler-IN: You don't have to throw here, but it doesn't hurt anything.

The patch will show as skipped if Applied is false. You can log the error directly instead of throwing using the methods on the submodule. These will probably be moved off to a dedicated or injected logger pattern at some point.

wonkotron:
Aserai Cheap Caravans are fixed as of e1.3.0.226834, should this PR still be merged?

Tyler-IN:
Does this patch apply over top of 1.3.0 incorrectly? If not, we still want it for backwards compat, if it does we need to add checks to IsApplicable so that it returns false.

Tyler-IN:
We can test in an RC

Issue #155, opened 2020-04-15
Trampler doesn't work.


Skau:
Riding perk.
Effect: `+50% charge damage`

PR #217, opened 2020-04-23
Trampler implementation

JoeFwd:
(#155)
PR implementing the Trampler perk.

When a horse charge is detected, the ```ChargeDamageCallback(Agent attackerAgent, ...)``` method is called.

This one internally calls ```GetAttackCollisionResults(Agent attackerAgent, ...)```.
This method prints the damage to the combat log. So, we have to update the horse charge damage before it is being printed. The easiest way to do that is to postfix the function call calulating the horse charge damage.

This specific method is ```GetAttackCollisionResults ~Overloaded (BasicCharacterObject AttackerCharacter, ...)```. Unfortunatly, the ```AttackerCharacter``` argument is wrongly initialised for a horse rider resulting to a null reference. 

Therefore, this PR uses a transpiler to patch ```GetAttackCollisionResults(Agent attackerAgent, ...)``` by getting the correct rider's character. This enables us to check in the postfix if the character has the Trampler perk or not and to update the damage output.


src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch.cs
```diff
@@ -0,0 +1,134 @@
...
+      }
+      
+      CommunityPatchSubModule.Error($"{typeof(TramplerPatch).Name}: Could not find the starting point to add new instructions in {TargetMethodInfo}." + Environment.NewLine);
+      return codes.AsEnumerable();
+    }
+    
+    private static bool HeroHasPerk(BasicCharacterObject character, PerkObject perk)
+      => (character as CharacterObject)?.GetPerkValue(perk) ?? false;
+
+    private static int TramplerDamageModifier(int baseDamage) => baseDamage + baseDamage/2;
```
Tyler-IN, line 123:
```suggestion
    private static int TramplerDamageModifier(int baseDamage) => (int) MathF.Round(baseDamage * 1.5f);
```
Is this the same except with rounding?

src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch.cs
```diff
@@ -0,0 +1,134 @@
...
+      .GetMethod("UpdateCorrectCharacterForHorseChargeDamage", BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.DeclaredOnly);
+    
+    private static readonly List<LocalVariableInfo> characterLocalVariableInfos = TargetMethodInfo
+      .GetMethodBody().LocalVariables.Where(var => var.LocalType == typeof(BasicCharacterObject)).ToList();
+    
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions) {
+
+      if (characterLocalVariableInfos.Count != 2) {
+        CommunityPatchSubModule.Error($"{typeof(TramplerPatch).Name}: Expected two BasicCharacterObject  local variable instances in the original method."
```
Tyler-IN, line 95:
```suggestion
        CommunityPatchSubModule.Error($"{nameof(TramplerPatch)}: Expected two BasicCharacterObject  local variable instances in the original method."
```

src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch.cs
JoeFwd: Yeah it is the same, although rounding is not necessary in this context since we will always have a bonus integer of X or X + 0.5 which in an integer division will be X. But, if the bonus is bound to change, then rounding might prevent us an calculation error.

JoeFwd: Alright, let's do this as you mentionned.

src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch.cs
```diff
@@ -0,0 +1,136 @@
...
+      CommunityPatchSubModule.Error($"{nameof(TramplerPatch)}: Could not find the starting point to add new instructions in {TargetMethodInfo}." + Environment.NewLine);
+      return codes.AsEnumerable();
+    }
+    
+    private static bool HeroHasPerk(BasicCharacterObject character, PerkObject perk)
+      => (character as CharacterObject)?.GetPerkValue(perk) ?? false;
+
+    private static int TramplerDamageModifier(int baseDamage) {
+      var tramplerDamage = baseDamage * (1 + (decimal) DefaultPerks.Riding.Trampler.PrimaryBonus);
+      return tramplerDamage - Math.Truncate(tramplerDamage) >= 0.5m ? (int) Math.Ceiling(tramplerDamage) : (int) Math.Floor(tramplerDamage);
```
Tyler-IN, line 124: What the heck is this! :sweat_smile:
```suggestion
      var tramplerDamage = baseDamage * (1.0f + DefaultPerks.Riding.Trampler.PrimaryBonus);
      return tramplerDamage - (int) MathF.Round(tramplerDamage, MidPointRounding.AwayFromZero);
```
I think you meant to get the fraction by using truncate, but truncate keeps the integer part and gets _rid_ of the remainder or fraction...

Also, let's not use the `decimal` type here. :shit:

JoeFwd: ```tramplerDamage - Math.Truncate(tramplerDamage)``` gets the decimal part. An example would be 8.5 - 8 = 0.5.

The ```MathF``` library doesn't apply to ```NetStandard 2.0``` but to ```NetStandard 2.1```. (https://docs.microsoft.com/en-us/dotnet/api/system.mathf?view=netcore-3.1#moniker-applies-to)

Taleworlds implemented their own ```MathF``` and ```Mathf``` libraries. Neither of these have a Round method with a MidPointRounding parameter.

We can use the ```Math``` library and casting ```tramplerDamage``` to a double without losing any precision:
```csharp
private static int TramplerDamageModifier(int baseDamage)
  => (int) Math.Round((double) (baseDamage * (1.0f + DefaultPerks.Riding.Trampler.PrimaryBonus)), MidpointRounding.AwayFromZero);
```

Or use Taleworlds MathF library :
```csharp
private static int TramplerDamageModifier(int baseDamage) {
  var tramplerDamage = baseDamage * (1.0f + DefaultPerks.Riding.Trampler.PrimaryBonus);
  return tramplerDamage - (float)MathF.Floor(tramplerDamage) >= 0.5f ? MathF.Ceiling(tramplerDamage) : MathF.Floor(tramplerDamage);
}
```

Tyler-IN: Truncate does not get the decimal part.

https://dotnetfiddle.net/pl8PkV

Using System.Math will be fine.

JoeFwd: I still cannot understand how you came to this conclusion. 😄 

Yes, it is perfectly clear that Truncate does not get the decimal part but the integer part.

```tramplerDamage - Math.Truncate(tramplerDamage)``` returns the fraction.  It is evaluated before being consumed by the ```<=``` function. 

Anyway, I will commit the first suggestion later on.

Comment (TYoung86) [reply src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch.cs:124]: Ah I am blind, I was missing the `tramplerDamage - ` as part of the operation.

Comment (TYoung86) [reply src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch.cs:124]:
```cs
return tramplerDamage - (float)Math.Round(tramplerDamage, MidPointRounding.AwayFromZero);
```

TYoung86: And I'm posting with the wrong account again 🥴

Tyler-IN: [Here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbhr0mrTj36CRGMRKm4WADQAcSObUbN23PgJYBhCPgAOvADYwoAZVcA3KTFk15AMyKSEykDAAyMNhQXACCwBAcGGyJXAAm4gDmsekAYlDYYHwQEgwA3jQMVQyV1QDaALIwGAAWEGlCDk4AFE2t7Z32TgDy9sUSLLGZmbC4kp4wo5a8AF7Y4wCUALq1VcRBzCEAZk4Q6wwp3GmxAO7YAJ75dgBarhDdJ2cYDAgb5bvVKq8I4MbpuXAwFCsIS4Nwcez2aAYGBpDYAwGA4gAdgY4MhSlSaQAKhAAHJRWYYdQwTKuNxgbBOaLojGs7oANRgRWg5FI+hssHWMHpjOiAFUJNgjjBur8NiwSSKmVBuhtjKy9jjLulbg8nvhXlAILlGU5gIUANaytUAgC+AIBjWabQ6XV6zoGXSWvBKOimM1880WY34q3WPq4LI11UIDD6LsGIxDvsm01mQZETnEWW2AP2wQYn3O2uud0eRoNbxNTjNlo+p3Ov3+1Gj2Lj6xauXYhNlaDjvDSiPEyUJWUmZf1hogNpb1Xts6qjvjnqG7v6rqG3pT/vTvAW3vwYfGUejDFjy43SfGfrTgb3MEz2a4mVzC6YB3Ixwb3xLuvLLyrU1zTAC1SHrL4fj+Co30xHFwPWDYGg7bsrl7ftBwgYcSzHP9JzeGdAXnQElw9S81wTL1kwmHc733ENDzWTYdjffNDkLb8GHyQpxng74m2g6NgVBPEoWEWF4URKBkVRE9YJ+Bg4FxCFRNyU5oCVZkYNPKoOS5DAeT5AUomRDSoAlXApRlOUFQgUzVXVDU2y4ooI2rWsQOtByqiI6oSPXRNyJXK8IxvAM5nvA8jwjWSNXPUjEy3ajb3ChZHy4HNmMxD8vwg5zxjc4CrSLPioJPNsEAYABSBhyC8hgfMXN8nX8t0LwSqjQt3OjlkY6KtO0uKWs3DrUzCjMuCzdLn1fLKC2KziChckoCstMD5v4sqcQqxTeL+JDWgVKBuAZZFPLtB0mra1r4so69Rq64MevDEoYtZQaKOGu6aJSh8JqfF9MuqVjP3Y3LFvyoDLQCXjIObVstoU9tWi7Ikjq4E6rIIud/GoW0gA=)'s me idly toying with rounding to see what the best options are.

Considering switching the mod over from Net Standard 2.0 to Net Fw 4.7.2

iPherian: @Tyler-IN very interesting! those sse41 instructions are nice.

Tyler-IN: I'll make the change in dev.

PR #218, opened 2020-04-23
Early story visible timeout patch alterations

Tyler-IN:
Primarily changes behavior for when the const is missing.

Implicitly refers to StoryMode, as we do for SandBox.


iPherian:
Pretty cool! I'd have taken much longer, i'm not very good at reflection.

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
```diff
@@ -2,112 +2,123 @@
...
       // defer our updates until some time has passed, because they depend on whether 
       // FirstPhase or SecondPhase is active, and story Phase information is only 
       // updated after all OnQuestStarted handlers have fired.
-      new NextHourlyTickListener(new Action<QuestBase>(SetStoryVisibleTimeoutIfNeeded), quest);
+      onHourlyTickBoxed.Value = () => {
+        SetStoryVisibleTimeoutIfNeeded(quest);
+        CampaignEvents.HourlyTickEvent.ClearListeners(onHourlyTickBoxed.Value);
+      };
+
+      CampaignEvents.HourlyTickEvent.AddNonSerializedListener(this, onHourlyTickBoxed.Value);
```
iPherian, line 119: IIRC, the parameter to `ClearListeners` needs to be the same as the first param to the corresponding `AddNonSerializedListener` call

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
```diff
@@ -2,112 +2,123 @@
...
+      // give OnHourlyTick's closure a late bound self-ref
+      var onHourlyTickBoxed = new StrongBox<Action>();
+
       // defer our updates until some time has passed, because they depend on whether 
       // FirstPhase or SecondPhase is active, and story Phase information is only 
       // updated after all OnQuestStarted handlers have fired.
-      new NextHourlyTickListener(new Action<QuestBase>(SetStoryVisibleTimeoutIfNeeded), quest);
+      onHourlyTickBoxed.Value = () => {
+        SetStoryVisibleTimeoutIfNeeded(quest);
+        CampaignEvents.HourlyTickEvent.ClearListeners(onHourlyTickBoxed.Value);
```
Tyler-IN, line 116:
```suggestion
        CampaignEvents.HourlyTickEvent.ClearListeners(this);
```
So it'd be `this`?

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
Tyler-IN: See below.

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
iPherian: yes 👍

iPherian:
Had a lot of suggestions, so it was easier to put them in a commit. You can revert / tweak as you prefer. Will add notes momentarily.

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
```diff
@@ -31,10 +31,10 @@ public sealed class EarlyStoryVisibleTimeoutPatch : IPatch {
 
     private static object SecondPhaseInstance => SecondPhaseInstanceGetter.Invoke(null, null);
 
-    private static readonly int FirstPhaseTimeLimitInYears = ExtractFirstPhaseTimeLimitInYears();
-
     private static readonly FieldInfo FirstPhaseTimeLimitInYearsField = AccessTools.Field(FirstPhaseCampaignBehaviorType, "FirstPhaseTimeLimitInYears");
 
+    private static readonly int? FirstPhaseTimeLimitInYears = ExtractFirstPhaseTimeLimitInYears();
+
```
iPherian: Reversed the order of these because `ExtractFirstPhaseTimeLimitInYears()` depends on `FirstPhaseTimeLimitInYearsField` already being set.

iPherian: Also `FirstPhaseTimeLimitInYears` is nullable to allow the rest of the code to not do anything if we can't retrieve the value.

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
```diff
@@ -54,18 +54,23 @@ public sealed class EarlyStoryVisibleTimeoutPatch : IPatch {
...
-          .GetValue(FirstPhaseTimeLimitInYearsField.IsStatic ? null : FirstPhaseInstance);
+      catch {}
+      try {
+        if (FirstPhaseTimeLimitInYearsField.IsStatic) {
+          return (int) FirstPhaseTimeLimitInYearsField.GetValue(null);
+        }
       }
+      catch {}
+
+      return null;
```
iPherian: To avoid throwing if we can't retrieve the value as either a constant or a static and just return null.

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
```diff
@@ -82,14 +87,14 @@ private static bool IsFirstStoryPhase()
         && SecondPhaseInstance == null;
 
     private static void SetStoryVisibleTimeoutIfNeeded(QuestBase quest) {
-      if (!IsFirstStoryPhase() || !quest.IsSpecialQuest || !quest.IsOngoing)
+      if (FirstPhaseTimeLimitInYears == null || !IsFirstStoryPhase() || !quest.IsSpecialQuest || !quest.IsOngoing)
```
iPherian: Avoid changing the quests at all if we couldn't get the time limit value. Otherwise it would have been 0 and we would have insta-cancelled all the quests.

Tyler-IN:
Try the other branch, does it work?

Tyler-IN:
IsRemainingTimeHidden returns true for some (all?) of the quests.

iPherian:
Same problem with the old PR branch.

> IsRemainingTimeHidden returns true for some (all?) of the quests.

Ahh.

Tyler-IN:
According to QuestItemVM, days remaining should still show up if the quest 'IsOngoing' - idk what that entails other than maybe being active?

iPherian:
Looking at QuestItemVM, I think it would be hidden merely if QuestBase had the hidden property.

```
this.IsRemainingDaysHidden = !this.Quest.IsOngoing || this.Quest.IsRemainingTimeHidden;
```

It seems to be an abstract property set per quest class. For example, in the "talk to lords" quest, which is class `StoryMode.Behaviors.Quests.FirstPhase.BannerInvestigationQuestBehavior.BannerInvestigationQuest` in `Modules\StoryMode\bin\Win64_Shipping_Client\StoryMode.dll`:

```
internal class BannerInvestigationQuest : StoryModeQuestBase
    {

    ...

      public override bool IsRemainingTimeHidden
      {
        get
        {
          return true;
        }
      }
```

iPherian:
The classes we want are inside the namespace `StoryMode.Behaviors.Quests.FirstPhase`. I suppose the algo would go something like this, if reflection is capable of this:

```
For every class inside StoryMode.Behaviors.Quests.FirstPhase {
    if (the class inherits from CampaignBehaviorBase) {
        look for any nested classes which inherit from StoryModeQuestBase, for each {
            Harmony patch the IsRemainingTimeHidden property to always return false
        }
    }
}
```

src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
iPherian: Reverted this part.

iPherian:
@Tyler-IN Strongly suggest that we merge this one for release. It fixes a previous PR which is already in dev, including turning a hard dependency on StoryMode into a lazy dependency.

PR #223, opened 2020-04-23
add perk Briber from roguery

miguelcjalmeida:
This pull request adds the perk Briber from roguery. It increases the chances of bribery dialogues to appear by 15%. When you try to force a villager or caravan into giving you their money, it opens bribery related dialogues in case of success. The chances are defined like this: When your party is stronger than the villager/caravan in a certain proportion, you get 66% chance of success; otherwise you receive only 33%. So the perk increases this chance, opening bribery dialogues on success.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Cunning/Roguery/BriberPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Cunning/Roguery/BriberPatch.cs
@@ -1,0 +1,177 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using HarmonyLib;
+using Helpers;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.SandBox.CampaignBehaviors;
+using TaleWorlds.CampaignSystem.SandBox.CampaignBehaviors.VillageBehaviors;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Cunning.Roguery {
+
+  public sealed class BriberPatch : PatchBase<BriberPatch> {
+    
+    public override bool Applied { get; protected set; }
+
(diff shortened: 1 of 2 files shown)


src/CommunityPatch/Patches/Perks/Cunning/Roguery/BriberPatch.cs
```diff
@@ -0,0 +1,177 @@
...
+    // ReSharper disable once RedundantAssignment
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static bool PrefixCaravansSurrender(ref bool __result) {
+      __result = Bribe(7, .1f, .1f);
+      return false;
+    }
+    
+    private static bool Bribe(int randomIndex, float phaseOneAcceptablePowerRatio, float phaseTwoAcceptablePowerRatio) {
+      var isLogicalSurrender = PartyBaseHelper.DoesSurrenderIsLogicalForParty(MobileParty.ConversationParty, MobileParty.MainParty, phaseOneAcceptablePowerRatio);
+      var bribeSuccessChances = isLogicalSurrender ? 67 : 33;
```
Tyler-IN, line 163: Put some comments around this, 67 is 2/3rds rounded up, 33 is 1/3rd rounded down?

Tyler-IN: Versus your description in the PR, note the subtle differences;
When your party is stronger than the villager/caravan in a certain proportion, you get 67% chance of success. Default is 34%. (?)
Otherwise you receive only 33%. Default is 15%. (?)

Tyler-IN: Is there some way to grab these defaults and add perk.PrimaryBonus to them?

Tyler-IN: Is there a way to extract these defaults somewhere?

Tyler-IN: Looks like not really  - you have them hashed, so this is good enough.

PR #225, opened 2020-04-23
add perk Negotiator and Bribe master from roguery

miguelcjalmeida:
This PR adds two perks that reduces bribe prices by 50% each. However, I made they stack multiplicatively so there will always be a price for a bribe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ This is going to be a mod that just fixes up some things in Mount &amp; Blade 2:
     * Eye for Loot
     * For the Thrill
     * Slip into Shadows
+    * Negotiator
   * Two Handed
     * Quick Plunder
     * Eviscerate
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ This is going to be a mod that just fixes up some things in Mount &amp; Blade 2:
     * For the Thrill
     * Slip into Shadows
     * Negotiator
+    * Bribe Master
   * Two Handed
     * Quick Plunder
     * Eviscerate
--- a/src/CommunityPatch/Patches/Perks/Cunning/Roguery/BribeMasterPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Cunning/Roguery/BribeMasterPatch.cs
@@ -1,0 +1,61 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.SandBox.GameComponents.Map;
+using TaleWorlds.Core;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Cunning.Roguery {
+
+  public sealed class BribeMasterPatch : PatchBase<BribeMasterPatch> {
+    
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo TargetMethodInfo = typeof(DefaultBribeCalculationModel).GetMethod(nameof(DefaultBribeCalculationModel.GetBribeToEnterLordsHall), BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
+    private static readonly MethodInfo PatchMethodInfo = typeof(BribeMasterPatch).GetMethod(nameof(Postfix), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.DeclaredOnly);
+
+    public override IEnumerable<MethodBase> GetMethodsChecked() {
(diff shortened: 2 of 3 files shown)


src/CommunityPatch/Patches/Perks/Cunning/Roguery/NegotiatorPatch.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(ref int __result) {
+      var scout = Hero.MainHero?.PartyBelongedTo?.EffectiveScout;
+      if (scout == null) return;
+
+      var perk = ActivePatch._perk;
+      if (!scout.GetPerkValue(perk)) return;
+
+      __result = (int) (__result * -perk.PrimaryBonus);
```
iPherian, line 62:
```suggestion
      __result = (int) (__result * Math.abs(perk.PrimaryBonus));
```

README.md
```diff
@@ -64,6 +64,7 @@ This is going to be a mod that just fixes up some things in Mount &amp; Blade 2:
     * Eye for Loot
     * For the Thrill
     * Slip into Shadows
+    * Negotiator
```
Tyler-IN, line 67: Bribe Master?

src/CommunityPatch/Patches/Perks/Cunning/Roguery/NegotiatorPatch.cs
Tyler-IN: What is the value from the primary bonus?

iPherian: > What is the value from the primary bonus?

`-0.5f`

src:

TaleWorlds.CampaignSystem.DefaultPerks.InitializePerks(), e1.1.2

```
// Negotiator perk
this.RogueryNegotiator.Initialize("{=lH8ZMjEo}Negotiator",
                                  "{=LLj618tA}Bribes are 50% cheaper.",
                                  DefaultSkills.Roguery,
                                  this.GetTierCost(6),
                                  this.RogueryBriber,
                                  SkillEffect.PerkRole.Scout,
                                  -0.5f,  // PRIMARY_BONUS
                                  SkillEffect.PerkRole.None,
                                  0.0f,
                                  SkillEffect.EffectIncrementType.AddFactor);
```

README.md
Tyler-IN: FYI, it's in the other PR.

src/CommunityPatch/Patches/Perks/Cunning/Roguery/NegotiatorPatch.cs
Tyler-IN: I doubt they'd ever want bribes to _cost more_ but the behavior of just using negation is _ok_.

If you don't want to use Math.Abs / MathF.Abs here, you should verify that perk.PrimaryBonus <= 0 in IsApplicable.

iPherian: Well, sometimes the perks seem to be additive factors and sometimes multiplicative. Sometimes even in contradiction to the `AddFactor` argument. For example, Bribe master and negotiation both describe "lowering bribes by 50%" but Bribe Master's PrimaryBonus is 0.5 and Negotiators is -0.5. They're also both `AddFactor` type. Or maybe it's a vanilla bug.

Anyway he just multiplies it in the patch and the result shouldn't be negative so I thought Math.Abs would be best.

miguelcjalmeida: I really liked the idea of Math.Abs. This will make the solution shorter and understandable

src/CommunityPatch/Patches/Perks/Cunning/Roguery/BribeMasterPatch.cs
```diff
@@ -0,0 +1,61 @@
...
+      if (Applied) return;
+      CommunityPatchSubModule.Harmony.Patch(TargetMethodInfo, postfix: new HarmonyMethod(PatchMethodInfo));
+      Applied = true;
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(ref int __result) {
+      var perk = ActivePatch._perk;
+      if (!Hero.MainHero.GetPerkValue(perk)) return;
+      __result = (int) (__result * perk.PrimaryBonus);
```
iPherian, line 58: I suppose this as well

```suggestion
      __result = (int) (__result * Math.Abs(perk.PrimaryBonus));
```

iPherian:
You can choose whether to keep that last one, just seemed like a good idea.

PR #227, opened 2020-04-24
add perk slave trader from roguery

miguelcjalmeida:
Adds the roguery perk 'Slave trader' which increases your profits from ransoming by 20%. It seems every ransom related action calls the method CharacterObject.PrisonerRansomValue to determine the value of the prisioner; even when bartering them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,8 @@ This is going to be a mod that just fixes up some things in Mount &amp; Blade 2:
     * Eye for Loot
     * For the Thrill
     * Slip into Shadows
-    * Slave Trader
     * Merry men
+    * Slave Trader
   * Two Handed
     * Quick Plunder
     * Eviscerate


src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlaveTraderPatch.cs
```diff
@@ -0,0 +1,82 @@
...
+using HarmonyLib;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Cunning.Roguery {
+
+  public sealed class SlaveTraderPatch : PatchBase<SlaveTraderPatch> {
+
```
iPherian, line 13:
```suggestion
    private const float BonusRansomGoldPercent = .20f;

```

src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlaveTraderPatch.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+      );
+
+      _perk.Initialize(
+        textObjStrings[0],
+        textObjStrings[1],
+        _perk.Skill,
+        (int) _perk.RequiredSkillValue,
+        _perk.AlternativePerk,
+        _perk.PrimaryRole, .20f,
```
iPherian, line 62:
```suggestion
        _perk.PrimaryRole, BonusRansomGoldPercent,
```

src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlaveTraderPatch.cs
Tyler-IN: I think we can do this pattern in bulk to many perks in a refactor PR later.

src/CommunityPatch/Patches/Perks/Cunning/Roguery/SlaveTraderPatch.cs
Tyler-IN: I think we can do this pattern in bulk to many perks in a refactor PR later.

PR #228, opened 2020-04-24
add the roguery perk Merry men

miguelcjalmeida:
This PR adds the roguery perk  Merry men' that should allow you to recruit bandits. Recruiting bandits was already possible but would required a massive army to intimidate them. Not just that, it would require also luck. By having the perk, the recruitment should require just a minimum number of soldiers and nothing else. 

An important note. The perk ID was not unique, it was the same as the bow perk 'Merry Men'.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Cunning/Roguery/MerryMenPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Cunning/Roguery/MerryMenPatch.cs
@@ -64,6 +64,7 @@ namespace CommunityPatch.Patches.Perks.Cunning.Roguery {
       intimidated = PartyBaseHelper.DoesSurrenderIsLogicalForParty(
         MobileParty.ConversationParty, 
         MobileParty.MainParty, 
+        // requires that the bandits be merely less than 100% of our strength to surrender.
         1f);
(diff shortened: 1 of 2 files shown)


src/CommunityPatch/Patches/Perks/Cunning/Roguery/MerryMenPatch.cs
```diff
@@ -0,0 +1,72 @@
...
+      => ApplyPerkToMakeRecruitable(ref __result);
+
+    private static bool ApplyPerkToMakeRecruitable(ref bool intimidated) {
+      var perk = ActivePatch._perk;
+      if (MobileParty.MainParty.LeaderHero?.GetPerkValue(perk) != true) return true;
+      
+      intimidated = PartyBaseHelper.DoesSurrenderIsLogicalForParty(
+        MobileParty.ConversationParty, 
+        MobileParty.MainParty, 
+        1f);
```
iPherian, line 67: I'm posting like 5 million nitpicks right now, but a comment would be good here 😊

```suggestion
        // requires that the bandits be merely less than 100% of our strength to surrender.
        1f);
```

PR #230, opened 2020-04-24
add Companion Cavalry perk from tactics tree

miguelcjalmeida:
This PR adds the perk Companion Cavalry which increases cavalry morale damage by 10%, if leader has the perk active.

Morale loss is applied when a soldier dies in combat, applying a AOE morale damage to nearby friends. So the implementation checks if the killer has a mount and a leader with the perk active, then the it applies a second AOE morale damage, but just with the bonus amount (10%).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ This is going to be a mod that just fixes up some things in Mount &amp; Blade 2:
   * Two Handed
     * Quick Plunder
     * Eviscerate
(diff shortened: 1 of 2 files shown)


iPherian:
@Tyler-IN If github could give collaborators custom titles we could give @miguelcjalmeida  one called "MasterOfPerks" 😊

PR #236, opened 2020-04-25
EarlyStoryVisibleTimeoutPatch: check for game type being story mode

iPherian:
Instead of the patch stopping if we are in multiplayer, instead stop if we are in any game mode other than CampaignStoryMode.

I'm not absolutely certain, but basically because all the base game code checks in the this way (for CampaignStoryMode) before accessing story mode vars. Also hypothetically TW might add a new game mode in the future which is not multiplayer but is also not CampaignStoryMode and in that case the patch would incorrectly apply itself.

### DIFF
--- a/src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
+++ b/src/CommunityPatch/Patches/EarlyStoryVisibleTimeoutPatch.cs
@@ -18,6 +18,7 @@ namespace CommunityPatch.Patches {
 
     private static readonly byte[] IsRemainingTimeHiddenGetterBodyIl = {
       // e1.1.2.226306
+      // i.e. IL for "return true"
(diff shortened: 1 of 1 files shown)


iPherian:
Examples of the base game checking for `(game.GameType is CampaignStoryMode)`

in `StoryMode.StoryModeSubModule`

```csharp
protected override void OnGameStart(Game game, IGameStarter gameStarterObject) {
    if (!(game.GameType is CampaignStoryMode))
        return;

    // do stuff related to StoryMode
}

// and similar checks in OnGameEnd and OnGameInitializationFinished
```

in `StoryMode.StoryMode`

```csharp
public static StoryMode Current {
    get {
        return !(Game.Current?.GameType is CampaignStoryMode gameType) ? null : gameType.StoryMode;
    }
}
```

PR #253, opened 2020-04-30
add perk Elusive from Tactics

miguelcjalmeida:
Add the perk 'Elusive' from Tactics tree. The perk should reduce the troop cost when leaving battles. I have made the buff apply when you are trying to break in through the besieged camp as well.


src/CommunityPatch/Patches/Perks/Cunning/Tactics/ElusivePatch.cs
```diff
@@ -0,0 +1,103 @@
...
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void SiegePostfix(ref int __result, MobileParty party)
+      => ApplyPerk(ref __result, party);
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void SacrificePostfix(ref int __result, BattleSideEnum battleSide, MapEvent mapEvent) {
+      var side = battleSide == BattleSideEnum.Attacker ? mapEvent.AttackerSide : mapEvent.DefenderSide;
+      
+      foreach (var sideParty in side.Parties.Where(x => x.MobileParty != null))
+        ApplyPerk(ref __result, sideParty.MobileParty);
```
iPherian, line 90: Idk this might be a bit overpowered. It seems that if e.g. 2 party leaders in the side have this perk, the men sacrificed for the entire side is reduced by 50% and then 50% again, even if they only command a small fraction of the side's total men.

Maybe it would make more sense to have the perk's reduction be applied once and scaled by the proportion of men that are led by a leader with the perk?

```suggestion
      __result -= __result * percentMenLeadByLeaderWithPerk * ( 1.0f - perk.PrimaryBonus )
```


( just did `( 1 - bonus )` because the description of the perk is 50% fewer men with a value of 0.5f. So if they someday change it to 30% fewer men I'd say they'd set the bonus to be 0.3f so we'd have to 1 - 0.3 before mult. Admittedly this is a guess. )

iPherian:
Otherwise looks good.

src/CommunityPatch/Patches/Perks/Cunning/Tactics/ElusivePatch.cs
Tyler-IN: Yeah it's OP when stacking.

iPherian: In addition maybe if the army's leader has the perk then the bonus could be gotten for the entire army regardless of the individual party leaders. That's how TW does it for the Assuring Presence perk (which is 20% less army cohesion decay but has the role PartyLeader).

But otherwise fall back on it being based on the percentage of men who are in qualifying parties? This is just speculative but that's how I would do it.

Example of Assuring Presence perk from `CalculateCohesionChangeInternal`:

```CSharp
if (!army.LeaderParty.HasPerk(DefaultPerks.Leadership.AssuringPresence))
        return;
float perkMultiplyFactor = DefaultPerks.Leadership.AssuringPresence.PrimaryBonus * 0.01f;
float cohesionIncrease = Math.Abs(
    cohesionChange.ResultNumber  // negative
        * perkMultiplyFactor);
cohesionChange.Add(cohesionIncrease, DefaultArmyManagementCalculationModel._assuringPresencePerkText);
```

iPherian: Possible implementation of my suggestion:

iPherian/MnB2-Bannerlord-CommunityPatch@25d19399894823bb0b54a3a15f23c83a98ed885c

Tyler-IN:
Still waiting on @miguelcjalmeida to have time to respond.

src/CommunityPatch/Patches/Perks/Cunning/Tactics/ElusivePatch.cs
miguelcjalmeida: I was taking a look in your implementation and it looks great. I'll give some test to make sure it is working fine, then I'll let you know.

miguelcjalmeida: There was a bug with the implementation where it wasn't subtracting the sacrifice reduction from sacrifice total, but that was a bug already in my implementation too. Also, I refactored your code just a little bit according to ReSharper suggestions. 

I did some test and this is working fine. Thank you so much iPherian for suggesting and comming up with the solution!

iPherian: Nice catch!

PR #256, opened 2020-05-01
add the perk Bait from tactics

miguelcjalmeida:
This PR adds the perk bait, which should increase reinforcement range by 30% (party member).

In order to do so, I have patched the method responsible for bringing the nearby troops (both ally and enemy) into battle. Since the range values were hardcoded in a very big method, I had the idea to make a postfix that would search for nearby parties a second time, but this time with the range increased.

The down side is that the scan is done twice, however, since this is just a player feature that triggers only when his/her battle is going to start, it shouldn't be a performance problem.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Cunning/Tactics/BaitPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Cunning/Tactics/BaitPatch.cs
@@ -1,0 +1,143 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using HarmonyLib;
+using Helpers;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Library;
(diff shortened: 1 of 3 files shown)


src/CommunityPatch/Patches/Perks/Cunning/Tactics/BaitPatch.cs
```diff
@@ -0,0 +1,143 @@
...
+        }
+      );
+
+      _perk.Initialize(
+        textObjStrings[0],
+        textObjStrings[1],
+        _perk.Skill,
+        (int) _perk.RequiredSkillValue,
+        _perk.AlternativePerk,
+        _perk.PrimaryRole, 30f,
```
iPherian, line 58: you mean 0.30f? 😉

miguelcjalmeida: PerkHelper required the perk value to be in that format. It doesn't work with factors unfortunately.

PR #257, opened 2020-05-01
add perk Trusted Commanders from tactics

miguelcjalmeida:
Adds the perk 'Trusted Commander', which reduces the cohesion cost of armies by 50%. The description doesn't mention the value, I just kept the one already specified in the DefaultPerks class. Since this is a party member bonus, I made it stack multiplicatively.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Cunning/Tactics/TrustedCommandersPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Cunning/Tactics/TrustedCommandersPatch.cs
@@ -1,0 +1,116 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using HarmonyLib;
+using Helpers;
+using TaleWorlds.CampaignSystem.SandBox.GameComponents.Map;
+using static System.Reflection.BindingFlags;
+using static CommunityPatch.HarmonyHelpers;
+
+namespace CommunityPatch.Patches.Perks.Cunning.Tactics {
+
+  public sealed class TrustedCommandersPatch : PatchBase<TrustedCommandersPatch> {
+
+    public override bool Applied { get; protected set; }
+
+    private static readonly MethodInfo TargetMethodInfo = typeof(DefaultArmyManagementCalculationModel).GetMethod("CalculateCohesionChange", Public | Instance | DeclaredOnly);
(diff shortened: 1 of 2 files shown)


Tyler-IN:
For some reason it kept telling me that I didn't approve this. Now it finally shows up for me that I approved it. Weird.

src/CommunityPatch/Patches/Perks/Cunning/Tactics/TrustedCommandersPatch.cs
```diff
@@ -0,0 +1,116 @@
...
+      
+      if (Applied) return;
+      CommunityPatchSubModule.Harmony.Patch(TargetMethodInfo, postfix: new HarmonyMethod(PatchMethodInfo));
+      Applied = true;
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static void Postfix(ref float __result, Army army, StatExplainer explanation = null) {
+      var perk = ActivePatch._perk;
+      var reduction = CalculatePerkReduction(army, perk);
```
iPherian, line 78: So it's multiplicative tapering off towards 100%? So if 1 party qualifies, 50% reduction for entire army, if two 75%, if three parties 87.5%? I feel like a PartyMember bonus should merely change whether each party qualifies for the perk, and the decision to then apply it to the entire army should be based on specific conditions.

My impression is that it would be more balanced to do it like this:
- If the army's leader party qualifies for the perk, apply it to the entire army's cohesion change. This is how TW does it for the 'Assuring Presence' perk which similarly reduces cohesion decay by 20%. 
- Else get the percentage of men in the army who are in parties with the perk, and scale it based on that. So if that's 20% then the final reduction is 50% * 20% = 10%.

Example of Assuring Presence perk (from `CalculateCohesionChangeInternal`):

```CSharp
if (!army.LeaderParty.HasPerk(DefaultPerks.Leadership.AssuringPresence))
        return;
float perkMultiplyFactor = DefaultPerks.Leadership.AssuringPresence.PrimaryBonus * 0.01f;
float cohesionIncrease = Math.Abs(
    cohesionChange.ResultNumber  // negative
        * perkMultiplyFactor);
cohesionChange.Add(cohesionIncrease, DefaultArmyManagementCalculationModel._assuringPresencePerkText);
```

Tyler-IN:
Still waiting on @miguelcjalmeida to have time to respond.

src/CommunityPatch/Patches/Perks/Cunning/Tactics/TrustedCommandersPatch.cs
miguelcjalmeida: I have pushed the changes according to your suggestions, could you review again please?

iPherian: Looks good!

PR #279, opened 2020-05-09
Trampler perk implementation for 1.4.0 beta

JoeFwd:
Taleworlds added in the 1.4.0 beta patch the ```AttackInformation``` class in order to have clearer method signatures.
This change impacts the  ```GetAttackCollisionResults``` method with a new ```AttackInformation``` parameter holding most of the previous attack parameters (method signature before 1.4.0) especially the attacker character object.
This one is needed to check if the rider's character has got the trampler perk or not.

Moreover, the ```TramplerPatch2``` will not be loaded in patches prior to 1.4.0 since the compiler does not know what  the ```AttackInformation``` class is.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch2.cs
+++ b/src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch2.cs
@@ -1,0 +1,127 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Core;
+using HarmonyLib;
+using TaleWorlds.Engine;
+using TaleWorlds.MountAndBlade;
+using static System.Reflection.BindingFlags;
+using static CommunityPatch.HarmonyHelpers;
+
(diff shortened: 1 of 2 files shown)


src/CommunityPatch/Patches/Perks/Endurance/Riding/TramplerPatch2.cs
```diff
@@ -0,0 +1,127 @@
...
+    public override bool? IsApplicable(Game game) {
+      
+      if (OverloadedTargetMethodInfo == null)
+        return false;
+      
+      if (AlreadyPatchedByOthers(Harmony.GetPatchInfo(OverloadedTargetMethodInfo)))
+        return false;
+
+      if (!OverloadedTargetMethodInfo.MakeCilSignatureSha256().MatchesAnySha256(OverloadedTargetHashes))
+        return false;
```
iPherian, line 68: if it's more convenient, you can combine these two checks into a call to `IsTargetPatchable(...)` which does the same (by importing `PatchApplicabilityHelper`)

iPherian: oh, nvm.

PR #287, opened 2020-05-11
Adjust to changes in Bannerlord's Town api

iPherian:
A mid release update for 1.4 beta on 2020-05-11 removed the Town.All property so use Campaign.AllTowns instead.

This assumes that AllTowns represents specifically towns as opposed to towns OR castles. Probably true since there is also an AllCastles property.

### DIFF
--- a/src/CommunityPatch/Patches/Perks/Intelligence/Steward/RulerPatch.cs
+++ b/src/CommunityPatch/Patches/Perks/Intelligence/Steward/RulerPatch.cs
@@ -6,7 +6,7 @@ using HarmonyLib;
 using TaleWorlds.CampaignSystem;
 using TaleWorlds.Core;
 using static System.Reflection.BindingFlags;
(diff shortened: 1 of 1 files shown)


TYoung86:
This needs to be backwards compatible with other branches.

Do the previous branches still refer to this as All?
If so, we may need a helper that selects All or AllTowns getters by reflection now.

Tyler-IN:
This needs to be backwards compatible with other branches.

Do the previous branches still refer to this as All?
If so, we may need a helper that selects All or AllTowns getters by reflection now.

iPherian:
@Tyler-IN previous game vers have it as All, and our previous usage was by All (though it was only this patch that used it apparently).

Changed to access AllTowns then All by reflection.

iPherian:
Changed to use Campaign.AllTowns

tested working (a town gives an extra companion limit) in 1.4 beta, 1.3.x and 1.2.x

PR #333, opened 2020-05-18
Bump perks to 1.4.1

JoeFwd:
I can confirm that these work on 1.4.1 and therefore assume that perks patching the same method as those will work aswell.
* Conroi - Bannerlord (Increased Party size)
* Prominence (Influence gain)
* Healthy Scout (Increased max hitpoints)
* Nourrish Settlement (Increased town prosperity)
* ProsperousReign (Increased village hearth)
* Trampler perk
* Menu when encountering an army feat
* Elusive Perk
* Nomadic traditions perk - Khuzait Cavalry feat (Confirmed by fnzr)
* Bait perk
* Companion cavalry perk
* Filled to the brim
* Scavenger perk
* Slave Trade
* Warmonger
* Eye For Loot - Raiding Party
* Land grants for veterans policy
* Heavier siege engines

Two handed perk patched in this project does not apply to 1.4.1 since Taleworlds revamped all of the perks in this category.


PraxisLP:
I managed to build and test this version. (some files needed to be excluded because a method was missing some arguments, and some true s needed to be changed to 0s but it built)

Seems to work a lot better than the version on Nexus. When can we expect an updated Nexus version?

PR #344, opened 2020-05-19
use correct perk in Roguery.MerryMenPatch

alejandro-y:
Currently Roguery.MerryMenPatch is referring to BowMerryMen perk, because it has the same name as RogueryMerryMen perk.
This RogueryMerryMen is indeed rogue as it doesn't even have a property in DefaultPerks so had to find it by StringId.
The workaround is quite ugly but I could not find any better way, maybe you can help with this?
Btw TW has also BowMountedArcher and RidingMountedArcher with the same problem.

### DIFF
--- a/src/CommunityPatch/Patches/PerkPatchBase.cs
+++ b/src/CommunityPatch/Patches/PerkPatchBase.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Runtime.CompilerServices;
 using CommunityPatchAnalyzer;
 using JetBrains.Annotations;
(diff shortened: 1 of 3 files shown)


fnzr:
I think we could change the PerkPatchBase._perk accessibility to protected, and then you override the Reset function with  `Reset() => _perk = PerkObject.FindFirst(x => x.StringId == "RogueryMerryMen")`.

We could also change the constructor argument from "Id" to "StringId", but that would require refactoring a bunch of stuff so I think the first option is better for the few perks that need it.

Tyler-IN:
`PerkPatchBase` should have a 2nd protected constructor that allows you to specify the perk finder `Func<PerkObject>`

One of the reasons to use the perk's localization ID is to ensure the perk is initialized if they change the way serialization works (and break something) - special cases are fine though I'd recommend the finder check both the StringId and Id.

alejandro-y:
Thanks guys, looks much better now.

PR #371, opened 2020-06-05
Skip patches when PatchObsolete/PatchNotBefore is used

JoeFwd:
BuilderPatch and BuilderPatch2 patches were wrongly applied. BuilderPatch is obsolete since e1.3.0 and was still applied after 1.3.0+.

- Skip patches that have the PatchObsolete / PatchNotBefore attribute and being not applicable to current game version.

- Makes skipped patches clearer with the addition of "Obsolete" or "Non compatible" patch messages instead of "Skipped".


JoeFwd:
1.4.2 breaks this PR.
I will need to adjust it to the new ApplicationVersion API.

